PR #49, opened 2013-11-25
New workflow

sickill:
Here's the idea for the new recording/publishing workflow:

* `asciinema rec` - record the terminal session, it gets saved in your local "repository",
* `asciinema ls` - see what's local (not published),
* `asciinema ls -r` - see what's remote (published),
* `asciinema rm <id>` - delete it,
* `asciinema play <id>` - replay it in your terminal,
* `asciinema edit <id>` - fire up $EDITOR and edit title, description and the content itself (fixing typos!),
* `asciinema process <id> ...` - post-process the recording (optimizations like: removing idle time, speeding it up, etc. Also, filtering the recorded content by running through arbitrary command) 
* `asciinema push <id>` - upload/publish to asciinema.org
* `asciinema search <text>` - search for asciicasts including given text (in title, description and content)

See this gist for example workflow: https://gist.github.com/sickill/c83e9f972c89dd76ef95

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,9 +27,22 @@ Arch Linux users can install the
 
 ## Usage
 
-Check the available commands and options with:
+Record:
 
-    asciinema -h
+    $ asciinema rec
+    d734ae3
+
+Replay:
+
+    $ asciinema play d734ae3
+
+Edit:
+
+    $ asciinema edit d734ae3
+
+Publish:
+
+    $ asciinema push d734ae3
 
 ## Authors
 


sickill:
`asciinema rec [<filename>]`, `asciinema play <filename>` and `asciinema upload <filename>` has been implemented in 1.0.0.rc1: https://github.com/asciinema/asciinema/releases/tag/v1.0.0.rc1

Instead of using "local repository" as described in the first comment, you just pass filename (or a path), which is simpler. Instead of having `asciinema edit` you can just open and modify recorded asciicast in your editor as it is now a JSON file (see [asciicast format](https://github.com/asciinema/asciinema/blob/master/doc/asciicast-v1.md)).

I'm closing this one. `process`, `rm` and `search` will be tackled separately.

PR #77, opened 2015-01-08
Add basic man(1) page

jakubjedelsky:
Every well behaved unix program should have own man page. This is basic one heavily inspired by your README.md file and contributing page.


sickill:
Awesome! I was thinking about man page but I never created one so I was postponing this again and again :)

PR #90, opened 2015-03-23
Man page should correspond to current commands

jakubjedelsky:
Edit some details and add info for play and upload commands.

### DIFF
--- a/man/asciinema.1
+++ b/man/asciinema.1
@@ -3,7 +3,7 @@
 asciinema \- Terminal recorder for asciinema.org service
 .SH "SYNOPSIS"
 .I asciinema
-[\-h] [\-v] command [<args>]
+[\-h] [\-\-version] command [<args>]
 .SH "DESCRIPTION"
 Command line client (terminal recorder) for asciinema.org service.
 .PP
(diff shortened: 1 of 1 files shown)


sickill:
I completely forgot about the man page :/
I have no knowledge of (g)roff unfortunately so this was always something on my to-do list that wast buried by dust. I wonder if there's some tool that could automatically generate man page out of for example markdown.

sickill:
Just found this: https://rtomayko.github.io/ronn/ronn.1.html
Not sure how useful it would be though.

man/asciinema.1
```diff
@@ -39,20 +35,33 @@ Recording finishes when you exit the shell (hit \fBCtrl+D\fP or type \fIexit\fP)
 \-c
```
sickill: One thing here: `-c` has now `--comand` alternative.

jakubjedelsky:
It will be cool to generate it automatically! But there are few things which, I think, shouldn't be in man page (install and build info). But the true is it would be much more simple. I don't know ronn tool, but can take care about man page..

sickill:
I was thinking about cutting out some part(s) of readme with something like awk, piping it to ronn. We can definitely look into that some day.

sickill:
Thanks! :heart:

PR #95, opened 2015-05-06
Simple patch to enable max-wait for playback

theatrus:
Take the -w option on playback, which reduces all intervals in the recorded
file to the specified number of seconds (or less)

### DIFF
--- a/asciicast/player.go
+++ b/asciicast/player.go
@@ -1,13 +1,14 @@
 package asciicast
 
 import (
+	"math"
 	"time"
 
 	"github.com/asciinema/asciinema/terminal"
 )
 
 type Player interface {
-	Play(string) error
+	Play(string, float64) error
 }
 
 type AsciicastPlayer struct {
@@ -18,14 +19,14 @@ func NewPlayer() Player {
 	return &AsciicastPlayer{Terminal: terminal.NewTerminal()}
 }
 
-func (r *AsciicastPlayer) Play(path string) error {
+func (r *AsciicastPlayer) Play(path string, maxWait float64) error {
 	asciicast, err := Load(path)
 	if err != nil {
 		return err
 	}
 
 	for _, frame := range asciicast.Stdout {
-		delay := time.Duration(float64(time.Second) * frame.Delay)
+		delay := time.Duration(float64(time.Second) * math.Min(maxWait, frame.Delay))
 		time.Sleep(delay)
 		r.Terminal.Write(frame.Data)
 	}
--- a/asciicast/player.go
+++ b/asciicast/player.go
@@ -6,8 +6,6 @@ import (
 	"github.com/asciinema/asciinema/terminal"
 )
 
-const playbackDefaultMaxWait = 3600.0
-
 type Player interface {
 	Play(string, uint) error
 }
--- a/asciicast/player.go
+++ b/asciicast/player.go
@@ -31,7 +31,7 @@ func (r *AsciicastPlayer) Play(path string, maxWait uint) error {
 	if adjustedMaxWait <= 0 {
 		adjustedMaxWait = playbackDefaultMaxWait
 	}
-	
+
 	for _, frame := range asciicast.Stdout {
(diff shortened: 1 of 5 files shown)


sickill:
Thanks. Adding some comments inline.

asciicast/player.go
```diff
@@ -18,14 +19,14 @@ func NewPlayer() Player {
...
-func (r *AsciicastPlayer) Play(path string) error {
+func (r *AsciicastPlayer) Play(path string, maxWait float64) error {
 	asciicast, err := Load(path)
 	if err != nil {
 		return err
 	}
 
 	for _, frame := range asciicast.Stdout {
-		delay := time.Duration(float64(time.Second) * frame.Delay)
+		delay := time.Duration(float64(time.Second) * math.Min(maxWait, frame.Delay))
```
sickill: If `maxWait` is 0 (default value of float64) then this will always `time.Sleep(0)`.

commands/play.go
```diff
@@ -12,6 +12,6 @@ func NewPlayCommand() *PlayCommand {
 	}
 }
 
-func (c *PlayCommand) Execute(filename string) error {
-	return c.Player.Play(filename)
+func (c *PlayCommand) Execute(filename string, maxWait uint) error {
+	return c.Player.Play(filename, float64(maxWait))
```
sickill: Can we pass it as uint here, to make it consistent with `RecordCommand#Execute`?

main.go
```diff
@@ -127,9 +127,10 @@ func main() {
 		err = cmd.Execute(command, title, assumeYes, maxWait, filename)
 
 	case "play":
+		maxWait := uintArg(args, "--max-wait", cfg.RecordMaxWait())
```
sickill: I'm not sure if it's good to use `cfg.RecordMaxWait()` here as a default. `RecordMaxWait` is a setting for recording, maybe it would be better to have separate one for playback, like `PlayMaxWait`? Given that you can replay other people's files I can imagine a scenario where I want to use `maxWait = 2` for all playbacks but don't use it for my recordings. Or vice versa. It's hard to imagine what would be the usage patterns here, but it's important to note that `maxWait` applied to recording is a "destructive" operation (it happens once and its effect is persisted) and applied to playback is "non destructive".

What do you think? Any other ideas?

theatrus: More than happy to make a new cfg. option for this, and I agree making it default to a reasonable value (say, 2s) makes sense.

theatrus:
Updated for comments, new config option and checking for 0-sleeps.

asciicast/player.go
```diff
@@ -18,14 +21,19 @@ func NewPlayer() Player {
...
 	}
 
+	adjustedMaxWait := float64(maxWait)
+	if adjustedMaxWait <= 0 {
+		adjustedMaxWait = playbackDefaultMaxWait
+	}
+
 	for _, frame := range asciicast.Stdout {
-		delay := time.Duration(float64(time.Second) * frame.Delay)
+		delay := time.Duration(float64(time.Second) * math.Min(adjustedMaxWait, frame.Delay))
```
sickill: I'd rather not modify delay at all when maxWait is 0. Here you modify it when it's > 3600. 

What about something like this:

```go
delay := frame.Delay
if maxWait > 0 && delay > maxWait {
  delay = maxWait
}
time.Sleep(time.Duration(float64(time.Second) * delay))
```

It's similar to what happens during recording: https://github.com/asciinema/asciinema/blob/master/asciicast/stream.go#L41-L45

theatrus:
Matched logic from recording

sickill:
Looks good, thanks!

laughedelic:
It looks like the limitation of setting this parameter as whole secods is artificial. I'd like to set this option to a value less than a second, so either setting it in milliseconds or as fractional seconds (I see `float64` in the code).
@sickill I guess the change is easy and I can submit a PR if you're okay with it.

sickill:
There's no reason to not support fractions of a second. As long as I'll still be able to pass 2 instead of 2.0 I'm happy to accept PR. Note that we use this same option for both `play` and `rec` commands so it should be updated in both places.

PR #102, opened 2015-05-23
Add environment variable ASCIINEMA_CONFIG_HOME for customizing path to the config dir

zmwangx:
Currently the config path `$HOME/.asciinema/config` is hard-coded, which is not so cool for users who keep all their config stuff in `$HOME/.config` for a cleaner `HOME` and easier backup (following the [XDG Base Directory Specification](http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html)). Therefore, I'm introducing an `ASCIINEMA_CONFIG` environment variable in this PR to let users keep their config files wherever they want. (Personally I would set it to `$HOME/.config/asciinema/config`.)

### DIFF
--- a/util/cfg.go
+++ b/util/cfg.go
@@ -81,15 +81,36 @@ func GetConfig(env map[string]string) (*Config, error) {
 }
 
 func loadConfigFile(env map[string]string) (*ConfigFile, error) {
-	homeDir := env["HOME"]
-	if homeDir == "" {
-		return nil, errors.New("Need $HOME")
+	pathsToCheck := make([]string, 0, 4)
+	if env["ASCIINEMA_CONFIG_HOME"] != "" {
+		pathsToCheck = append(pathsToCheck,
+			filepath.Join(env["ASCIINEMA_CONFIG_HOME"], "config"))
+	}
+	if env["XDG_CONFIG_HOME"] != "" {
+		pathsToCheck = append(pathsToCheck,
+			filepath.Join(env["XDG_CONFIG_HOME"], ".asciinema", "config"))
+	}
+	if env["HOME"] != "" {
+		pathsToCheck = append(pathsToCheck,
+			filepath.Join(env["HOME"], ".config", "asciinema", "config"))
+		pathsToCheck = append(pathsToCheck,
+			filepath.Join(env["HOME"], ".asciinema", "config"))
 	}
 
-	cfgPath := filepath.Join(homeDir, ".asciinema", "config")
+	cfgPath := ""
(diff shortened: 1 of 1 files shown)


sickill:
I like that. It's not cool indeed to have it hard coded.

But I think it would be even better to have the env var specify asciinema's config directory instead of a file. Config loading could then check the following locations in this order:

* `$ASCIINEMA_CONFIG_HOME/config`
* `$XDG_CONFIG_HOME/asciinema/config`
* `$HOME/.config/asciinema/config`
* `$HOME/.asciinema/config` (to be backwards compatible)

Having `ASCIINEMA_CONFIG_HOME` as a path to a directory would be more future proof (allowing for example for storing more files there, like certificates, keys).

What do you think?

zmwangx:
That's the best solution. I intentionally made it simple before because not every maintainer is happy with the idea of XDG, but now that you've laid out the full solution, I'm more than happy to oblige.

Here's what I'm doing in the updated commit:

* First, check the following locations in order (skip if the corresponding environment variable is undefined or empty):

  * `$ASCIINEMA_CONFIG_HOME/config`
  * `$XDG_CONFIG_HOME/asciinema/config`
  * `$HOME/.config/asciinema/config`
  * `$HOME/.asciinema/config`

  Use the first one that is found.

* If not found, create a new config file at the first location listed above whose corresponding environment variable is nonempty; i.e.,

  * Use `$ASCIINEMA_CONFIG_HOME/config` if `ASCIINEMA_CONFIG_HOME` is
    nonempty;
  * Next, use `$XDG_CONFIG_HOME/asciinema/config` if `XDG_CONFIG_HOME`
  is nonempty;
  * Finally, use `$HOME/.config/asciinema/config` if `HOME` is nonempty.

* If none of the above works, exit with error `"Need $HOME"`.

Disclaimer: I seldom deal with Go, so please excuse me if my code is not very Go-like.

sickill:
@zmwangx this is perfect. I have nothing add

sickill:
I've made a small fix here: https://github.com/asciinema/asciinema/commit/7a5e25e846ab5949b8f415a4252bcb1697227cb9

Thanks!

zmwangx:
Oops sorry about the typo. My pleasure to contribute to this great tool.

zmwangx:
By the way, the ["configuration file" section](https://github.com/asciinema/asciinema#configuration-file) section of README needs an update to reflect this change.

sickill:
Yeah, I'll update README, thanks for the reminder!

PR #111, opened 2015-06-23
Support playing file from stdin

Jellyfrog:
This would be good for piping, for example:
`$ curl http://example.com/asciinema-recording.gz | gunzip | asciinema play -`

Never used Go before, but seems to work.


sickill:
Looks great, thanks!

sickill:
Thanks to this patch it's now possible to:

    curl -sL https://asciinema.org/a/20055.json | asciinema play -

Jellyfrog:
Ah, I was looking for a url to the original json!

Thanks for merging!

PR #112, opened 2015-06-24
add build on FreeBSD

swills:


sickill:
Thanks. Last time I've been building FreeBSD binaries on Mac or Linux they were segfaulting. Have you had a chance to test if these work correctly when compiled on non-FreeBSD system?

sickill:
I just saw this: http://src.mouf.net/ports/revision?rev=390467 . I'm not that familiar with Ports and the packaging process, does it mean it is already in Ports now? If so, the instructions for installing are `sudo port selfupdate && sudo port install asciinema` ?

Jellyfrog:
You can install it from ports with: `cd /usr/ports/textproc/asciinema && make install`
With pkg you can use `pkg install asciinema` (but it's not in pkg yet)

swills:
What @Jellyfrog says is correct. The packages should appear in a few days. I setup a dev environment on Mac and the FreeBSD amd64 and i386 binary produced works fine.

sickill:
@swills great, thanks for the report. I have added ports installation method to the site: https://asciinema.org/docs/installation (and mentioned FreeBSD support on the homepage).

swills:
I think it would be better to document the `pkg install asciinema` method. Also, are you going to merge this pull request or does it seem redundant? I think it would be good to allow both methods.

sickill:
I didn't add pkg install because you said it's not available via pkg yet. Also, I'm not that familiar with package instalation / release process. Feel free to send a PR when/if you think it's good idea to add it.

PR #114, opened 2015-06-26
allow installing on FreeBSD

swills:


sickill:
Thanks. Let me build and upload FreeBSD binaries first, then I'll merge it.

PR #143, opened 2016-02-04
Give live feedback on terminal size

colemickens:
This gives the user immediate feedback while they're resizing their terminal.

It works better than it looks here, tmux doesn't play nice with asciinema-inception: (see demo in subsequent comment). ~~https://asciinema.org/a/5jbmd6en6p3swoij9s1u2cjxe (also you can ignore most of the latter half of it, I got confused and tried to upload the outer file while I was still inside the outer asciinema session)~~

### DIFF
--- a/asciicast/recorder.go
+++ b/asciicast/recorder.go
@@ -11,8 +11,8 @@ import (
 )
 
 const (
-	optimalCols = 120
-	optimalRows = 30
+	warnCols = 120
+	warnRows = 30
 )
 
 type Recorder interface {
@@ -30,7 +30,6 @@ func NewRecorder() Recorder {
 func (r *AsciicastRecorder) checkTerminalSize() chan<- bool {
 	rows, cols, _ := r.Terminal.Size()
 	doneChan := make(chan bool)
-	optimalSize := fmt.Sprintf("%s%dx%d%s", "\x1b[0m\x1b[32m", optimalCols, optimalRows, "\x1b[0m\x1b[33m")
 	go func() {
 		winch := make(chan os.Signal, 1)
 		signal.Notify(winch, syscall.SIGWINCH)
@@ -46,10 +45,7 @@ func (r *AsciicastRecorder) checkTerminalSize() chan<- bool {
 				if cols != newCols || rows != newRows {
 					cols, rows = newCols, newRows
 					currentSize := fmt.Sprintf("%dx%d", cols, rows)
-					if cols == optimalCols && rows == optimalRows {
-						currentSize = fmt.Sprintf("%s%dx%d%s", "\x1b[0m\x1b[32m", cols, rows, "\x1b[0m\x1b[33m")
-					}
-					util.ReplaceWarningf("Current terminal size is %s. Optimal is %s.", currentSize, optimalSize)
+					util.ReplaceWarningf("Current terminal size is %s.", currentSize)
 				}
 			case <-doneChan:
 				return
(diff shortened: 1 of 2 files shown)


colemickens:
Better demonstration: https://asciinema.org/a/4ocz1iaq6dsdnuu98doq5629v

sickill:
Good stuff! I have one suggestion about detecting change of terminal size. Terminal sends `WINCH` signal to current process (to `asciinema` during recording) and we can handle this signal instead of polling.

Here's an example of how this is already used in other part of the recorder: https://github.com/asciinema/asciinema/blob/master/terminal/terminal.go#L47-L56

Would you mind updating the PR to use signals?

colemickens:
@sickill updated, thanks!

sickill:
One thing I noticed is that when you resize the terminal to a small size and then resize it back to max size there is some weirdness with the lines. See the screenshot from iTerm: https://dl.dropboxusercontent.com/u/12119988/asciinema-rec-oops.png . This looks like the result of overwriting the line from the beginning (`\r`).

Re the "optimal size", the way you implemented it suggests there is a "golden" size one should aim at. While 120x30 is a good size, all the smaller sizes are equally good - the smaller the size the better chance it will be viewable in full on most devices. My concern is, people could feel they actually need to resize their smaller terminals until they get "green ok" and one of my goals for this project is it should be "no hassle recording".

Anyway, the "optimal size" thing is about the wording (and possibly change then condition to check if the size is `<=` instead of `==` to optimal one), we can tweak it easily. I wonder though if there's anything we can do about the line weirdness from the screenshot. Are you experiencing this too? What terminal app are you using?

colemickens:
I've updated it and removed the notion of optimal. The messaging is more like it was originally and there's no green hinting at 120x30.

I can't reproduce the line effect you're seeing there. I'm testing with termite and tmux is always running. It looks like maybe that method of carriage return is not portable. I'll try to see if something else works, but I don't have a mac to test with.

sickill:
I have tested on both iTerm 2 and Terminal.app and I can see this effect on both. I will fire up an Ubuntu VM and check on xterm, and few others.

colemickens:
Here's it working with `xterm`: https://www.youtube.com/watch?v=Q1hnP_vwI54

I can try it with others under Linux. I've started the process of getting a Mac but that could take an undefined amount of time.

I tried searching "iTerm2 carriage return" but haven't come up with much.

colemickens:
Hi @sickill. I have pull this PR and built it on El Capitan. Using iTerm2 (Build 2.9.20160313), I am unable to reproduce the behavior you're showing, unfortunately.

colemickens:
Just tried in Terminal.app and can't repro there either.

sickill:
Interesting. I don't think that matters but which shell did you tried it in? I'm using zsh.

colemickens:
I'm using `zsh` as well. Note the iTerm I used was a pre-release version, but I'm surprised we're seeing different behavior even in the default Terminal.app with zsh. I even removed tmux from my testing so that it wouldn't be interfering.

sickill:
Hey @colemickens!

We have just released asciinema v1.3.0, which is based on original Python implementation of asciinema (v0.9.8 codebase). I really appreciate your work on this and I'm sorry it wasn't merged yet 🙈 . Given the current impl is Python I won't be able to merge this PR into master, but I have merged it into `golang` branch which we'll keep.

colemickens:
Anywhere I can read about the transition back to python? Motivation/pitfalls of Go/etc?

sickill:
http://blog.asciinema.org/post/and-now-for-something-completely-different/

Issue #121, opened 2015-07-12
[Feature Request]compress data for upload


dlintw:
My upload bandwidth is limited.
Could the upload command send by compressed data instead of json text format?

sickill:
How big is the file you try to upload?

dlintw:
Well, 5MB on 64k upload bandwidth..

Jellyfrog:
Gziping the output would yield a >90% compression ratio.
Maybe its a good idea to always compress the files, even for local saved files?

sickill:
FYI #146 was reverted as it wasn't working.

bkil:
Related, but not limited to upload transfer encoding (and with potential lossiness): #515

PR #146, opened 2016-02-12
Gzip data for upload

bentranter:
Relevant issue: #121

This tiny PR gzips the body of the POST request before uploading it, and adds the appropriate `Accept-Encoding` header to the request. As far as I can tell, no change needs to be made on the Rails end, since `Net::HTTP` knows to automatically decompress gzipped requests if they contain the `Accept-Encoding: gzip` header.

I tested this by compiling locally and uploading [an asciicast](https://asciinema.org/a/5a4ro8drmyekusbrzbrt3n5ur) and it seemed to work, but if there's more I should do to ensure this is okay, let me know.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"bytes"
+	"compress/gzip"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


sickill:
Hey! This looks good. I'll do a bit of testing (also against asciinema.org which has Nginx in front of Rails) and then merge. Thanks!

sickill:
@bentranter are you sure you tested this? This code is using `gzip.NewReader` which is a **decompressor**, not a **compressor**.

It just failed for me with error: `Error: gzip: invalid header`.

bentranter:
Oh man, what was I thinking... yeah using a `gzip.Reader` doesn't make any sense. This _did_ work for me, but I just realized I _didn't change the import path after I forked and cloned this_, so I didn't actually test my own changes...  ugh I'm super sorry @sickill, this was a very dumb mistake on my part.

I've looked into doing this properly, and it seems that the server doesn't like gzipped requests. If I send the request,

```http
POST /api/asciicasts HTTP/1.1
Host: asciinema.org
Authorization: Basic <token_goes_here>
Content-Type: application/x-gzip
Content-Encoding: gzip
User-Agent: asciinema/1.1.1 gc/go1.5.2 darwin-amd64
Content-Disposition: form-data; name="asciicast"; filename="testupload"
```

It responds with `500`. I've got the Rails end set up locally now, so I'll keep looking into this.

sickill:
No worries ;) I suspected that you might have tested using old binary. 

Anyway, content-encoding is right but you should leave content-type as it was originally. Then I think it should work.

Issue #136, opened 2015-12-16
Windows support request


vicneanschi:
Will there be a version for Windows?
Thanks

sickill:
See https://github.com/asciinema/asciinema/issues/47

svnpenn:
https://github.com/asciinema/asciinema/issues/292

svnpenn:
https://github.com/asciinema/asciinema/pull/152

Issue #47, opened 2013-10-29
Windows not supported, right?


maphew:
A quick pip install test from the Github repo reveals that asciicinema doesn't work out of the box on Windows, even though it does install an .exe to the Python Scripts folder. On my Win7 machine It fails with `ImportError: No module named termios`.

A little searching reveals there (probably) isn't a  Windows equivalent,
http://stackoverflow.com/questions/933745/what-is-the-windows-equivalent-to-the-capabilities-defined-in-sys-select-h-and-t, and that therefore asciicinema is likely not supposed to work on Windows.

It would be good if the Readme states what operating systems asciicinema is intended for / known to work on.

noisy:
@sickill is it possible to support cygwin?

sickill:
@noisy cygwin doesn't support UNIX-like pseudo-terminal so it's unlikely to make it work on Windows. All UNIX family systems (Linux, *BSD, Mac OS X), have PTYs, but Windows is a completely different OS.

masaeedu:
@sickill What about https://github.com/rprichard/winpty?

sickill:
@masaeedu feel free to experiment with it. I'm fine with someone maintaining a friendly Windows fork of asciinema. However, at this moment I'm interested in merging any Windows specific changes into the main repository. [asciicast format](https://github.com/asciinema/asciinema/blob/master/doc/asciicast-v1.md) is documented, so as long at recording on Windows produces compatible asciicast file, the web player should be able to replay it.

sickill:
@masaeedu There is Windows fork by @mattn - https://github.com/mattn/asciinema/tree/windows

sickill:
I was wrong saying "it will never support Windows" :)
It seems it now installs and works perfectly fine under Win10 (WSL): https://asciinema.org/a/123255

amistele:
March 28 2018, 
I tried a pip install of asciinema on Windows 10, and encountered the termios error. Is this still resolvable, or where do i go to fork a compatible version?

mattn:
@amistele see https://github.com/asciinema/asciinema/issues/150

svnpenn:
https://github.com/asciinema/asciinema/issues/136

svnpenn:
https://github.com/asciinema/asciinema/issues/292

svnpenn:
https://github.com/asciinema/asciinema/pull/152

mikemaccana:
[...]

PR #152, opened 2016-03-15
Windows Support

mattn:
channel2

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -44,6 +44,14 @@
 		{
 			"ImportPath": "golang.org/x/text/transform",
 			"Rev": "07b9a78963006a15c538ec5175243979025fa7a8"
+		},
+		{
+			"ImportPath": "github.com/mattn/go-colorable",
+			"Rev": "07b9a78963006a15c538ec5175243979025fa7a8"
+		},
+		{
+			"ImportPath": "github.com/mattn/go-isatty",
+			"Rev": "07b9a78963006a15c538ec5175243979025fa7a8"
 		}
 	]
 }
--- a/asciicast/asciicast.go
+++ b/asciicast/asciicast.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
+	"runtime"
 	"strings"
 
 	"github.com/asciinema/asciinema/Godeps/_workspace/src/golang.org/x/net/html"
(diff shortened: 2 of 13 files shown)


mattn:
Sorry, some of dependencies are entried as code.google.com/p/. AFAIK, it's gone. So I can't update Godep.json.

sickill:
I don't have a Windows box ready to test it at the moment, but will get one soon.

Anyway, I don't see any Windows specific PTY recording code. Does `kr/pty` package just work on Windows console??

mattn:
Ah, sorry, I forgot to add file. I will add soon.

> Does kr/pty package just work on Windows console??

No. it doesn't.

mattn:
done.

sickill:
Hey @mattn. This is hell of a PR :)

Looks like you put a lot of effort into emulating PTY under Windows. There's a lot going on in this code.
I suspected that there's no way to get a real PTY and real ANSI escape codes on Windows and the only way is to "scrape" the terminal buffer in regular interval.

I'm not a Windows user and I don't often have access to Windows so it would be very hard for me to maintain this (making sure all is okay, releasing .exe). Because of this and the fact it hasn't been widely tested yet, I'm hesitant to merge it at this point. I expect some inconsistency with ANSI codes, also scraping terminal buffer every 10ms is potentially CPU intensive (but I understand we don't know a better way!).

However, I really appreciate the work you have put in this. I'd be very happy to point people wanting to record on Windows to your fork. I can link to it from the README and also from asciinema.org docs. You could then maintain it at your own pace there, as a semi-official Windows fork.

What do you say?

mattn:
I don't mind that suggestion. Do it your best. :+1:

sickill:
Btw, I heard someone was able to run original asciinema on Windows 10 recently (thanks to bash/linux layer included in Windows).

svnpenn:
https://github.com/asciinema/asciinema/issues/292

svnpenn:
https://github.com/asciinema/asciinema/issues/136

svnpenn:
https://github.com/asciinema/asciinema/issues/47

haraldreingruber:
I nice workaround is installing Asciinema via WSL, start a recording session form WSL Bash and then open a CMD session via `cmd.exe`. 

It's a bit less convenient than being able to record from CMD directly but form an output perspective the only disadvantage is that the initial `cmd.exe` call is also recorded, and would need to be removed in a post-processing step in order to avoid confusion. Maybe somebody knows a trick to delay the recording so that capturing is started after calling `cmd.exe`?

rollingmoai:
This port for Windows works great: https://github.com/Watfaq/PowerSession-rs

PR #156, opened 2016-05-15
Changed max-wait parameter type to float64

laughedelic:
Nothing sophisticated, just changed the type in all places. I'm not proficient in Go, so excuse me if I did something stupid. Possible improvements:
- indicate this change in the docs explicitly
- check that the input parameter is positive and print an error otherwise


sickill:
Looks good, thanks!

PR #171, opened 2016-07-26
Feature: add review before uploading option

alexyorke:
This adds an option to review an asciinema before uploading (and adding the option to press the "r" key during the "enter to upload, ctrl-c to cancel" stage to review it.)

I just read the contributing guidelines now, and I would be happy to discuss it over IRC (I'm a little bit late for that though.)

Credit: a user who said a command chain like `asciinema rec && asciinema play && asciinema upload` is something that could be used to review a asciinema before it is uploaded.

### DIFF
--- a/asciinema/commands/record.py
+++ b/asciinema/commands/record.py
@@ -5,7 +5,8 @@ import tempfile
 from asciinema.commands.command import Command
 from asciinema.recorder import Recorder
 from asciinema.api import APIError
-
+from asciinema.player import Player
+import asciinema.asciicast as asciicast
 
 class RecordCommand(Command):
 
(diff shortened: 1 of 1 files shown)


19Coley-Wayne79:
Been In This Game For Years It Made Me An Animal, True To This Shit That’s How I Handle.

Decagon:
@sickill The build doesn't complete because "No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself." since this feature introduces a delay (where the user has to press a keyboard shortcut) to continue to send the presentation. This means that the build process stalls, waiting for the user to enter in the keyboard command. Is there something that I could add to this code which would allow the automated build process to work correctly?

sickill:
Frankly I'm not sure if this is a very common need, or if this should be enabled by default (which as you figured breaks tests).

Given the above and the fact that it has a conflict I'll close this. Thanks anyway!

Issue #177, opened 2016-09-30
Add config option to disable automatic uploading of asciicasts


cbarcenas:
Currently, it's too easy for users to accidentally publish sensitive screencasts to asciinema.org: all you need to do is accidentally omit a filename argument and then press <Enter> at the upload confirmation prompt.

At many security-minded organizations, particularly those that regularly deal with sensitive information, the automatic upload feature of `asciinema rec` would turn asciinema into a data leakage risk. This may in turn prevent its use at such organizations.

I propose that a configuration option, `upload`, be added to the `[record]` section of the config file. This option, which would default to `yes`, could be used to control the automatic upload functionality of `asciinema rec`. If set to `no`, automatic upload functionality would be disabled and the filename argument to `asciinema rec` would become mandatory.

I have implemented this proposal and will submit a PR for review.

sickill:
As this is linked to #178 I'm closing it too.

AkihiroSuda:
@sickill Could you reopen this issue as @saxophonicle said?

sickill:
My comments in the referenced PR were never addressed, and I never heard back from cbarcenas. Thus both PR and issue were closed.

sickill:
If you'd like to completely prevent uploads today you can put this in the config file, as temporary workaround:

```ini
[api]
url = http://nope
```

rtpt-jonaslieb:
We're having the same issue as @cbarcenas: We deal with sensitive data on a daily basis and the risk of automatically uploading screen casts is so significant that we are currently refraining from using asciinema, even though it's a really useful tool.

Therefore I'd offer to implement a solution. I would suggest re-implementing #178 (re)based on the current state of the repository, i.e. a configuration option that can be used to disable uploading. Additionally, I would also have asciinema look for the configuration at `/etc/asciinema/config` such that out administrators can roll out a centralized solution.

@sickill Does that sound okay to you?

IAXES:
> 
> 
> this issue was never fixed it needs to be reopened. The PR it was linked to was closed without merging and so this issue is lost to the void.

I frequently use this tool for tutorials, examples, etc.; and it's pretty awesome. That being said, I too was very concerned about accidental data leakage (i.e. there are use cases where this tool couldn't be used for legal reasons, solely due to the **potential** for accidental data leakage by inexperienced users or a simple [misconfiguration problem](https://owasp.org/www-project-top-ten/2017/A6_2017-Security_Misconfiguration) resulting in sensitive data being leaked.

[...]

PR #178, opened 2016-09-30
Add config option to disable automatic uploading of asciicasts

cbarcenas:
Implements feature discussed in #177.

### DIFF
--- a/asciinema/__main__.py
+++ b/asciinema/__main__.py
@@ -21,7 +21,7 @@ def positive_float(value):
 
 
 def rec_command(args, config):
-    api = Api(config.api_url, os.environ.get("USER"), config.api_token)
+    api = Api(config.api_url, os.environ.get("USER"), config.api_token) if config.record_upload else None
     return RecordCommand(api, args.filename, args.command, args.title, args.yes, args.quiet, args.max_wait)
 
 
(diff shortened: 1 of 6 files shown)


asciinema/__main__.py
```diff
@@ -21,7 +21,7 @@ def positive_float(value):
 
 
 def rec_command(args, config):
-    api = Api(config.api_url, os.environ.get("USER"), config.api_token)
+    api = Api(config.api_url, os.environ.get("USER"), config.api_token) if config.record_upload else None
```
sickill: What do you think about moving `upload` boolean up from `record.py` to `rec_command()` in `__main__.py`, and also generating a default filename (in tmp) there, then passing it down to `RecordCommand`. For example:

```python
def rec_command(args, config):
    api = Api(config.api_url, os.environ.get("USER"), config.api_token)
    if args.filename == "":
        filename = _tmp_path()
        upload = config.record_upload
    else:
        filename = args.filename
        upload = False
    return RecordCommand(api, filename, args.command, args.title, args.yes, args.quiet, args.max_wait, upload)
```

asciinema/__main__.py
```diff
@@ -82,7 +82,7 @@ def main():
     parser_rec.add_argument('-w', '--max-wait', help='limit recorded terminal inactivity to max <sec> seconds (can be fractional)', type=positive_float, default=maybe_str(cfg.record_max_wait))
     parser_rec.add_argument('-y', '--yes', help='answer "yes" to all prompts (e.g. upload confirmation)', action='store_true', default=cfg.record_yes)
     parser_rec.add_argument('-q', '--quiet', help='be quiet, suppress all notices/warnings (implies -y)', action='store_true', default=cfg.record_quiet)
-    parser_rec.add_argument('filename', nargs='?', default='', help='filename/path to save the recording to')
+    parser_rec.add_argument('filename', nargs='?' if cfg.record_upload else None, default='', help='filename/path to save the recording to')
```
sickill: This change wouldn't be needed with the change I suggested in `__main__.py`.

asciinema/commands/record.py
```diff
@@ -22,7 +22,7 @@ def __init__(self, api, filename, command, title, assume_yes, quiet, max_wait, r
     def execute(self):
         if self.filename == "":
```
sickill: This whole `if/else` block also wouldn't be needed with the suggested change.

sickill:
Given that the comments haven't been addressed and the branch is now in conflict with master (a lot has changed in a year) I'm closing it. We're working on new major version now, so if still interested feel free to open new issue where we can discuss whether and how we implement this. Thanks!

Issue #174, opened 2016-08-01
asciinema 1.3 does not provide a wrapper


carnager:
Since asciinema is now installed in python's site-packages directory, the binary won't be in $PATH anymore. 

asciinema should probably provide a little wrapper to be placed in $PATH. I guess setup.py is the place to do this.

sickill:
setup.py has it configured: https://github.com/asciinema/asciinema/blob/master/setup.py#L29
In all systems I checked asciinema installation it was correctly creating a wrapper script.

fornwall:
When I installed from source with `python3.5 setup.py install`, the resulting installation did not have the asciinema command-line wrapper.

I noticed the warnings

    /usr/lib/python3.5/distutils/dist.py:261: UserWarning: Unknown distribution option: 'entry_points'
    /usr/lib/python3.5/distutils/dist.py:261: UserWarning: Unknown distribution option: 'install_requires'

I had to `pip3 install setuptools` - after that the installation correctly set up the bin/asciinema wrapper.

I don't know enough about python packaging to know if this is the expected behaviour or not.

sickill:
Does anyone know a way to make it generate a wrapper in a reliable way (also in the absence of setuptools)?

sickill:
This has been solved in #179, the fix will be part of the next release.

sickill:
1.4.0 has been just released, including fix (#179) by @jakubjedelsky.

PR #179, opened 2016-10-12
setup: distutils can't create wrapper in $PATH

jakubjedelsky:
Module `distutils` don't know options `install_requires` and
`entry_points`. The second one is important for creating `asciinema`
binary within $PATH.

This should "resolve" Issue #174.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,11 +3,7 @@ import sys
 if sys.version_info[0] < 3:
     sys.exit('Python < 3 is unsupported.')
 
-try:
-    from setuptools import setup
-except ImportError:
-    from distutils.core import setup
-
+from setuptools import setup
 import asciinema
 
 


sickill:
Frankly I copy pasted this `try/except ImportError` piece from somewhere, as I read it was a recommended way to support both setuptools and distutils 😊

Anyway, is distutils already bundled with all versions of Python 3? Wouldn't this patch break the installation for more systems than it is broken for now?

jakubjedelsky:
`distutils` is enough for python libs, but don't support scripts (or binaries or how to name it). This is the job for `setuptools`. I checked their [tests](https://github.com/pypa/setuptools/blob/master/tox.ini) and it looks they support python 3.3, 3.4 and 3.5. So only 3.2 can be affected. I vote for get it out :)

Anyway no one shouldn't be affected because who don't have setuptools, he/she don't have asciinema bin in path :)

sickill:
Ok, fair enough :) Thanks!

Issue #118, opened 2015-06-30
[Feature request] Add parameter `--start-at NNN` to `asciinema play`


KurtPfeifle:
While editing an ASCiinema file, it would be nice to be able to start a playback from a specific line or time into the recording. Example:

```
 asciinema play  --start-at 56s my.rec
 asciinema play  --start-at 3215l my.rec
```

would start the playback at 56 seconds after the original starttime, or at line no. 3215 of the JSON source file.

parvbhullar:
Would like see that feature.

sickill:
We now have #180 open for this.

PR #180, opened 2016-10-13
Add start-at option to play command

jakubjedelsky:
option -s / --start-at allows user to start replay of a file at a
specified sec. It's quite stupid implemetnation, but should be enough for
review of a big file.

The logic behind is in Player class where sum of delays runs first in
every loop and if this sum is bigger than intended time, than we start
replaying.

This is a part of solution of Issue #118.

### DIFF
--- a/asciinema/__main__.py
+++ b/asciinema/__main__.py
@@ -26,7 +26,7 @@ def rec_command(args, config):
 
 
 def play_command(args, config):
-    return PlayCommand(args.filename, args.max_wait)
+    return PlayCommand(args.filename, max_wait=args.max_wait, start_at=args.start_at)
 
 
 def upload_command(args, config):
@@ -88,6 +88,7 @@ For help on a specific command run:
     # create the parser for the "play" command
     parser_play = subparsers.add_parser('play', help='Replay terminal session')
     parser_play.add_argument('-w', '--max-wait', help='limit terminal inactivity to max <sec> seconds (can be fractional)', type=positive_float, default=maybe_str(cfg.play_max_wait))
+    parser_play.add_argument('-t', '--time', metavar='SEC', dest='start_at', help='start replay at <SEC>-th second', type=positive_float, default=0)
     parser_play.add_argument('filename', help='local path, http/ipfs URL or "-" (read from stdin)')
     parser_play.set_defaults(func=play_command)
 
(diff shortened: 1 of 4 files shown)


sickill:
Looks great. There's one thing I'd like to discuss here before merging.

I planned to add an option to specify playback speed (similar to `speed` option in web player). I was thinking about `-s / --speed` option, so we would have a name clash with `-s` here. 

Should we maybe reserve `-s` for speed, and use `-t` for start-at (similar to `?t=10` URL param on asciinema.org)? I'm open to other naming suggestions.

jakubjedelsky:
No prob. I didn't know about `t` parameter on the webservice so I used the first character I figure out :) Maybe it will be fine to change `--start-at` to `--time`. Or use `-f/--from` (but `-f` can be misleading to `filename`) and add `-t/--to` to be able to replay just a selected part of record.

sickill:
Yet another option would be to have a single parameter for from/to, for example `-t 10-20` would play from 10 to 20 secs. Right now it would only support `-t 10`, without range, and we could extend it in the future maybe. Not saying it should be named `t`, just thought about the idea of "merged" from/to option :)

jakubjedelsky:
Tadaaah, sorry for so long delay. `-t` option is there. Hope it won't be mixed up with `-t` for `rec` command.

sickill:
And sorry for no reply for long time :)

Looks good. We have a conflict but should be trivial to fix that.

Anyway, I was thinking a bit about the name for this opton.
**t** / **time** is okay-ish (may be confused with `rec`'s **t**, though title doesn't really have any real meaning here so maybe it's not that bad).
**s** / **s**tart-at would be great, but we use **s** for **s**peed now.
I was also thinking about **d** / **d**rop (as in "drop first 5 secs"), **s**kip (but **s** conflict again).
Yet another options is to use -S / --start-at. 

I don't see a clear winner (yet, and sorry for nitpicking!) :)

sickill:
I just realized ignoring first n sec of stdout is not enough. Terminals are state machines so to have correct playback we need to print everything from the start, just not sleep until we reach desired start time.

sickill:
Since this change doesn't work correctly (see my prev comment), and many files changed/conflicted since we discussed it, I'm gonna close it. Resolving the conflicts would probably take more effort than preparing the patch from scratch on the current code.

PR #181, opened 2016-10-13
respect PEP8 coding style

jakubjedelsky:
No new functions, just coding edit coding style to pass with pep8 check. Except `E501 line too long`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,12 @@ python:
   - "3.4"
   - "3.5"
 
-script: make test
+before_install:
+  - pip install pep8
+
+script:
+  - find . -name \*.py -exec pep8 --ignore=E501 {} +
+  - make test
 
 notifications:
   irc:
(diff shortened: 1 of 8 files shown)


sickill:
Thanks Jakub!

sickill:
Hmm, looks like this change is not compatible with #179.

sickill:
Moving this setuptools import to the top is an option, but it will result in "no setuptools found" error for people on Python 2, unless they have setuptools installed.

Another option would be to exclude `setup.py` from pep8 checks.

jakubjedelsky:
I'll check it in a free time today. Maybe it will be fine to have `setuptools` as a requirement for instalation. Anyway, `setuptools` is so widespread that almost everyone have to have it installed :)

sickill:
I'm gonna fully trust you on this as I have little experience with Python's packaging specifics and current state of Python ecosystem 😛

PR #182, opened 2016-10-14
pep8: inlude modules on top of a file

jakubjedelsky:
Repair tests broken by #179 and #181


jakubjedelsky:
Just moving it up.

Anyway, I think that checking, that user uses python > 3 is useless. It should be fine just to inform in readme or so.

sickill:
👍

Issue #185, opened 2016-12-16
Storage of asciinema API key in config file


samdmarshall:
I would like to include my asciinema config file as part of my dot files that i keep in a public repo; the only thing stopping me right now is that it would embed my api key as plaintext. Are there any plans to allow for alternative methods of storing this value, such as environment variable, that would allow for asciinema to consume it without having to store it in the clear?

Thanks!

sickill:
That's a nice idea!

samdmarshall:
@sickill I have opened a PR addressing this issue, now waiting on that to be merged.

sickill:
@samdmarshall thanks! I'll take a look shortly.

PR #187, opened 2016-12-28
adding support to loading API token from an environment variable

samdmarshall:
This addresses #185, by adding a new fallback method of an environment variable named `ASCIINEMA_TOKEN` to store the user's API token so that sensitive information doesn't have to be kept in plaintext on disk.

### DIFF
--- a/asciinema/config.py
+++ b/asciinema/config.py
@@ -33,7 +33,11 @@ class Config:
             try:
                 return self.config.get('user', 'token')
             except (configparser.NoOptionError, configparser.NoSectionError):
-                raise ConfigError('no API token found in config file')
+                environment_variable_fallback = self.env.get('ASCIINEMA_TOKEN')
+                if environment_variable_fallback is not None:
+                    return environment_variable_fallback
+                else:
+                    raise ConfigError('no API token found in config file, and ASCIINEMA_TOKEN is unset')
 
     @property
     def record_command(self):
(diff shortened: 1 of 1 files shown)


asciinema/config.py
```diff
@@ -33,7 +33,11 @@ def api_token(self):
             try:
                 return self.config.get('user', 'token')
             except (configparser.NoOptionError, configparser.NoSectionError):
-                raise ConfigError('no API token found in config file')
+                environment_variable_fallback = self.env.get('ASCIINEMA_TOKEN')
```
sickill: Can we call it `ASCIINEMA_API_TOKEN`, to be consistent with `api.token` from the config file, and with parallel `api.url, ASCIINEMA_API_URL`?

samdmarshall:
Yeah, that seems like a much cleaner change-set. I have made those changes and updated the review now.

asciinema/config.py
```diff
@@ -28,12 +28,12 @@ def api_url(self):
     @property
     def api_token(self):
         try:
-            return self.config.get('api', 'token')
+            return self.env.get('ASCIINEMA_API_TOKEN', self.config.get('api', 'token'))
```
sickill: I think this will raise exception if you set env var but don't have `api.token` in config file because `self.config.get('api', 'token')` will get evaluated first, before its value can be passed to `self.env.get(..)`.
Maybe using Python's `or` would do the job here?

samdmarshall:
OK, i tested it out and it seems ready to go.

sickill:
Looks great! Thanks for you contribution @samdmarshall!

Issue #97, opened 2015-05-08
[CLI] Change playback speed


vvv:
Ability to speed up / slow down `asciinema play` would be very useful.

[ttyrec](http://0xcc.net/ttyrec/index.html.en) has this feature:

```
TTYPLAY(1)                                                    TTYPLAY(1)

NAME
       ttyplay - player of the tty session recorded by ttyrec

SYNOPSIS
       ttyplay [-s SPEED] [-n] [-p] file

DESCRIPTION
       Ttyplay  plays the tty session in file, which was recorded previ-
       ously by ttyrec(1).

       When -p option is given, ttyplay output the file as it grows.  It
       means  that  you  can  see  the  "live"  shell session running by
       another user.

       If you hit any key during playback, it will go right to the  next
       character  typed.   This is handy when examining sessions where a
       user spends a lot of time at a prompt.

       Additionally, there are some special keys defined:

       + or f  double the speed of playback.

       - or s  halve the speed of playback.

       1      set playback to speed 1.0 again.

OPTIONS
       -s SPEED
              multiple the playing speed by SPEED (default is 1).

       -n     no wait mode.  Ignore the timing information in file.

       -p     peek another person's tty session.

SEE ALSO
       script(1), ttyrec(1), ttytime(1)

                                                              TTYPLAY(1)
```

sickill:
Yeah, that is something I was considering. It's not a top priority thoughat the moment, sorry. If you often replay in terminal then maybe just stick to `ttyrec`?

sickill:
Should we close this issue?

vvv:
I guess we should. Thank you!

Drugoy:
The ability to change playback speed interactively is far superior to setting it for the whole video on start up via a cli arg.

PR #190, opened 2017-01-23
add --speed parameter to play command

bastiaanb:
implements part of #97


sickill:
Great, thanks!

Just one thing. Could you also update README and man page with a description of this new option?

bastiaanb:
Ah, overlooked man page. Added (README was already done)

sickill:
Perfect. Thank you for your contribution!

PR #193, opened 2017-02-22
Trap more signals than just SIGCHLD

typerlc:
Properly write a recording file if asciinema is terminated with SIGHUP, SIGTERM, SIGQUIT.  This happens, for example, if the machine is rebooted during a recording session.


sickill:
Good idea, thanks!

Issue #198, opened 2017-03-24
Consider resizing terminal in `asciinema play`


sickill:
Given we know width/height of the terminal that was used during recording session, we can send `\e[18;24;80t` (replace 80 with actual width and 24 with height) to the terminal at the beginning of the playback.

This escape sequence is part of "Window manipulation" sequence family, described in [XTerm documentation](http://invisible-island.net/xterm/ctlseqs/ctlseqs.html).

It isn't supported by all terminal emulators, but it seems to work with XTerm, Gnome Terminal, Terminal.app (macOS).

sickill:
This could be a default behavior, with command line switch and config option to enabled/disable this.

mattdeckard:
Thanks for adding this, @sickill . I didn't realize the recorder was in a different repo.

Issue #108, opened 2015-06-22
`^[[?1;2c` printed to terminal when playing locally


sickill:
Terminal itself is putting these characters on vim's stdin (emulating user typing them) during the recording, so vim can read and interpret them. It's a way of asking the terminal for its parameters. During playback asciinema sends the same escape codes that were emmited by vim but doesn't read from stdin so they stay in your prompt after asciinema finishes playback.

A potential solution would be to read everything from stdin during playback and ignore it.
Reading from stdin during playback would be also a good opportunity to add pause/resume functionality (space bar).

mattdeckard:
I like the solution of reading from (and ignoring) stdin more than the solution I proposed in https://github.com/asciinema/asciinema.org/issues/239. It would also keep users from messing up the playback by accidentally hitting the keyboard during a playback.

PR #199, opened 2017-03-24
Ignore stdin during playback

mattdeckard:
Potential fix for #108 using `curses`.

Also has the side-effect of clearing the terminal at the beginning of the playback and restoring it afterward, which seems like a good thing.

It would be pretty easy to implement #198 on top of this as well, since retrieving and setting the terminal size are simple functions in `curses`

### DIFF
--- a/asciinema/player.py
+++ b/asciinema/player.py
@@ -1,5 +1,6 @@
 import curses
 
+
 class CursesWrapper:
     def __init__(self):
         self._stdscr = curses.initscr()
@@ -14,6 +15,7 @@ class CursesWrapper:
         # Flush stdin and clean up terminal
         curses.flushinp()
         curses.endwin()
+
 
 class Player:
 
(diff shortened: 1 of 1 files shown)


mattdeckard:
@sickill Sorry I'm probably spamming your inbox with all of this but I just discovered this tool and have already started using it a lot, and I just couldn't help try my hand at fixing this issue.

sickill:
@mattdeckard thanks for this. However, I found a slightly better solution in #200, which doesn't have a side-effect mentioned by you (I'd argue whether it's a good or bad thing), and also prevents users from messing up the playback by hitting the keyboard. I really appreciate your help though!

mattdeckard:
Cool. I think this one also prevented users from interrupting the playback but I like your solution too.

Clearing the terminal at the start of playback might make more sense in #198

sickill:
Ah, you're right, somehow missed that :/

PR #214, opened 2017-06-16
Add fallback for `JSONDecodeError`

### LINKED ISSUE #168: Exception when decoding JSON in `asciinema play https://......`
When playing from http and the URL doesn't point to either asciicast json file or html page with proper link tag there's exception thrown in Python 3.4.5:

```
AttributeError: 'module' object has no attribute 'JSONDecodeError'
```

It seems a fix like [this one](https://github.com/getsentry/raven-python/pull/208) should do the trick.

crazymerlyn:
Fixes #168.

### DIFF
--- a/asciinema/asciicast.py
+++ b/asciinema/asciicast.py
@@ -6,6 +6,10 @@ import urllib.error
 import html.parser
 from .stdout import Stdout
 
+try:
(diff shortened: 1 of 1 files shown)


sickill:
Perfect, thanks!

PR #227, opened 2017-09-16
Recording to asciicast v2 format

sickill:


sickill:
In addition to introducing new file format, this adresses #127 (thanks to this new format 😄).

Issue #216, opened 2017-07-17
Including `command` in asciinema recordings is a security problem


cemerick:
I discovered that if I provide a command e.g. `asciinema rec -c 'export API_KEY=...; ./script.sh' output.json`, then that command will be included verbatim in `output.json` in the `command` slot. Now that I know this, I can take care to put sensitive information (like API keys, URLs for private services, etc) in a script or scrub the output to null out the `command` attribute, but I'd like to suggest that `command` be included only when explicitly requested.

sickill:
Good catch. For some reason it never occured to me (and I am person who does care about privacy in general).

We can either disable it by default and add a switch + config file option like you suggested, or we can include `command` when recording to a file but drop it from JSON when uploading (both when uploading implicitly through `asciinema rec` or when doing it explicitly from local file through `asciinema upload file.json`).

cemerick:
`rec` time is really when it matters; especially for those that are self-hosting the player (like me), the assumption is that you can just drop the JSON file anywhere as-is.

sickill:
Fair enough. There's not _that_ much use for it anyway - we display it on the recording page in place of a title when a title was not explicitly set and the recording explicitly specified the command via `-c`.

cemerick:
In that case, it seems like you want exactly the reverse of what you suggested: exclude when writing the file, but include when uploading?

sickill:
This was solved in #234.

PR #234, opened 2017-09-17
Don't save user command, to avoid leaking secrets

sickill:
This solves #216.

Example of secret leak:

    asciinema rec -c 'export API_KEY=secret-stuff; ./script.sh'

The resulting asciicast would include `"command": "export API_KEY=secret-stuff; ./script.sh"`.

### DIFF
--- a/asciinema/asciicast/v2.py
+++ b/asciinema/asciicast/v2.py
@@ -107,9 +107,6 @@ class Recorder:
         if title:
             header['title'] = title
 
-        if user_command:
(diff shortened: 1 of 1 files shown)


sickill:
I believe collection of `command` should be at least disabled by default. This PR disables it completely. We could make it opt-in, but I don't think there are enough people who need that to justify config file option to enable it. Also, asciinema-player and asciinema-server don't really need that information.

Issue #222, opened 2017-09-04
asciicast v2: environment variable capture spec


sickill:
As mentioned here https://github.com/asciinema/asciinema/pull/196#issuecomment-288148352 the set of environment variables saved by `asciinema rec` in `env` map should be defined/controlled. 

What could work is to have a white-list of env variables saved under `env` key in the recording header. It could default to `SHELL,TERM` and you could override it either via command line switch when recording or via config file option.

Having this, if you also want to save `$USER`, `$HOSTNAME` and `$CUSTOM_STUFF` you could set the white-list to for `SHELL,TERM,USER,HOSTNAME,CUSTOM_STUFF` and have all the information that's useful to you.

sickill:
This has been implemented in #230 and will be released with the upcoming asciinema 2.0.

Issue #221, opened 2017-09-04
asciicast v2: stored color theme spec


sickill:
We're discussing version 2 of asciicast format in #196, and I think adding optional colors object to the header, containing text foreground/background color and the palette for all 8(16) colors would be a nice addition. It could be saved automatically by `asciinema rec` and used by web player. 

As @mrabault said in #11, we can use and update a technique of obtaining the colors, as long as the palette representation in asciicast header stays the same/compatible with the web player. As for the actual technique, let's keep it discussed in #11.

Anyway, what we can add to v2 header is the optional key named `colors` (or `theme` ?):

```json
{
  "version": 2,
  "width": 80,
  "height": 24,
  "theme": {
    "fg": "#d0d0d0",
    "bg": "#212121",
    "palette": "#151515:#ac4142:#7e8e50:#e5b567:#6c99bb:#9f4e85:#7dd6cf:#d0d0d0:#505050:#ac4142:#7e8e50:#e5b567:#6c99bb:#9f4e85:#7dd6cf:#f5f5f5"
  }
}
```

I took inspiration to use single string value for all 16 colors from [Terminator theme syntax](https://github.com/mbadolato/iTerm2-Color-Schemes/blob/master/terminator/Afterglow.config) which is compact and it's easy to write/read/parse for both humans and machines.

sickill:
This was added to the v2 spec in #233.

PR #236, opened 2017-09-22
asciinema 2.0

sickill:
Highlights:

- recording and playback of new asciicast v2 format - #196 #227 #228
  - draft of v2 format spec: https://github.com/asciinema/asciinema/blob/develop/doc/asciicast-v2.md
  - writing to disk in real-time, preventing losing sessions on crashes/reboots 
  - enabling recording and playing of very long asciicasts without involving lots of RAM
- (optional) stdin recording in order to display keys pressed overlay in web player
- storing current terminal color theme in the recording - #221 #233 (specific technique to be figured out, see #11)
- ability to capture extra environment variables - #222 #230
- saving timestamp of recording session - #229

See [2.0 milestone](https://github.com/asciinema/asciinema/milestone/5) for all the things we plan to include.


SeriousM:
The recent commits smell like gold plating! Is the release near?

sickill:
@SeriousM very near. I don't have anything major to add to 2.0, probably only documentation tweaks. Now, would be really awesome if someone could give it a ride, testing the new shiny commands and options. Could you try the latest version from `develop` branch?

KurtPfeifle:
@sickill wrote further above:

> *"Now, would be really awesome if someone could give it a ride"*

I'm about to record a few little tutorials (to publish on asciinema.org as well). I'll give the develop branch a workout with it. But two questions first:

1. Will asciinema.org be able to re-play the v2 already? Or will that have to wait?
1. Is there a way to convert v2 format to v1?

sickill:
@KurtPfeifle asciinema.org supports v2 already, both the API endpoint and the player version there.

As for converting v2 -> v1, it would be trivial to write a converter script, as it would be mostly about shoving event stream into the header and serializing it as single JSON object.

I'm curious, what's the use case for doing v2 -> v1? I'd rather see people converting v1 -> v2, as it has many benefits over v1.

KurtPfeifle:
@sickill:
I do not have any real use case for v2 -> v1 conversion. I asked just in case... in case I finally had a good v2 recording I like, but asciiinema.org could not replay it from the web. So given your statement about asciinema.org supporting v2 already, I'm happy.

KurtPfeifle:
@sickill:
Ha!, I can now think of a use case for v2 -> v1 conversion: I'd like to edit the cast file after it is done. v2 makes it definitely more difficult to do so -- v1 is much easier (for me, at least).

sickill:
What specifically makes editing v1 easier than v2 for you?

KurtPfeifle:
@sickill asked above:

> *"What specifically makes editing v1 easier than v2 for you?"*

I was completely wrong, sorry. Had been looking at the wrong examples...

Meanwhile I generated a few short v2 files myself, and it should be fine!

SeriousM:
Woohoo! Congratulations for the 2.0.0 merge!

Issue #249, opened 2017-12-09
[Feature request][v2] Please add a `--overwrite` parameter


KurtPfeifle:
I've started to give a workout to v2.

One smallish thing, which still is able to annoy me:

1. I want to record to an existing file. However, I do not want to append to it, but to overwrite it. I get a warning *`asciinema: /path/file.asc already exists, aborting`*. (This, of course, is a good thing to have -- we do not want to blindly overwrite files.)

1. That means I have to use another shell command (*`rm /path/file.asc`*) in between of test recordings. It would be more convenient if I could just add a `--overwrite` parameter to do that instead, keeping the existing *good* default to not overwrite existing files.

sickill:
I had a hunch this will be needed sooner or later :)

AMR-KELEG:
I would like to work on this issue.
What I plan to do is:
1) add new command line argument in __main__.py (overwrite flag)
2) use os.remove(filename) in record.py if file already exists and overwrite flag is used

KurtPfeifle:
Thanks, @AMR-KELEG & @sickill !

PR #251, opened 2017-12-12
Add --overwrite option to rec

AMR-KELEG:
Closes https://github.com/asciinema/asciinema/issues/249

Add overwrite option to overwrite the recording file if it exists.
The overwrite flag overrides the append option.

### DIFF
--- a/asciinema/commands/record.py
+++ b/asciinema/commands/record.py
@@ -22,6 +22,7 @@ class RecordCommand(Command):
         self.assume_yes = args.yes or args.quiet
         self.idle_time_limit = args.idle_time_limit
         self.append = args.append
+        self.overwrite = args.overwrite
         self.raw = args.raw
         self.recorder = raw.Recorder() if args.raw else v2.Recorder()
         self.env = env if env is not None else os.environ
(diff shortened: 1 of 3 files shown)


asciinema/commands/record.py
```diff
@@ -43,7 +44,12 @@ def execute(self):
                 self.print_error("can't write to %s" % self.filename)
                 return 1
 
-            if os.stat(self.filename).st_size > 0 and not append:
+            if os.stat(self.filename).st_size > 0 and self.overwrite:
+                self.print_info("removing the file at %s" % self.filename)
```
sickill: It would probably be just fine to not print anything here, as the overwrite has been explicitly requested. Also, the overwrite could be implemented as truncate operation on file, without actually removing it, and the fact we actually do remove the file completely is implementation detail ;)

What do you think?

AMR-KELEG: How can i truncate the file?
Should i use https://linux.die.net/man/1/truncate and add a new method in the __pty_recioder.py__ file?
From my current point of view, truncating a file is much more complicated.

sickill: I wasn't clear, sorry. I didn't mean we should truncate it, removing s fine. I meant we could remove this `print_info` line :)

PR #261, opened 2018-02-19
Use `encodebytes` instead of `encodestring` (deprecated since Python 3.1)

delirious-lettuce:
> ### base64.encodestring(s)
> Deprecated alias of encodebytes().
> ***Deprecated since version 3.1.***

* https://docs.python.org/3/library/base64.html#base64.encodestring


sickill:
Many thanks! 👍

PR #262, opened 2018-02-19
Swap TERM and SHELL in v2 file format example

anowlcalledjosh:
The v2 file format [example for `env`](https://github.com/asciinema/asciinema/blob/develop/doc/asciicast-v2.md#env) has the `TERM` and `SHELL` variables swapped:

```json
"env": {
  "SHELL": "xterm-256color",
  "TERM": "/bin/bash"
}
```

This commit fixes that by swapping them back.

### DIFF
--- a/doc/asciicast-v2.md
+++ b/doc/asciicast-v2.md
@@ -71,8 +71,8 @@ Example env:
 
 ```json
 "env": {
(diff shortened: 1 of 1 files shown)


sickill:
Good catch, thanks!

PR #294, opened 2018-04-08
Add a fallback when multiprocessing is unsupported

### LINKED ISSUE #271: Using multiprocessing.Queue does not work on Android
Trying to use multiprocessing.Queue (which asciinema v2 started to do) on a python compiled for Android does not work:

> ImportError: This platform lacks a functioning sem_open implementation, therefore, the required synchronization primitives needed will not function, see issue 3770.

The linked issue is this: https://bugs.python.org/issue3770

While python on Android is a niche OS, there is quite some interest in asciinema from [Termux](https://termux.com/) users. So I'm wondering if you're interested in replacing multiprocessing.Queue (at least as an alternative) to make asciinema work there?

fornwall:
On platforms (Android) where multiprocessing is not supported we now use threads instead. Fixes #271.

### DIFF
--- a/asciinema/asciicast/v2.py
+++ b/asciinema/asciicast/v2.py
@@ -4,7 +4,16 @@ import json
 import json.decoder
 import time
 import codecs
-from multiprocessing import Process, Queue
+
+try:
+    # Importing synchronize is to detect platforms where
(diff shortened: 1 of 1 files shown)


sickill:
Thanks!

PR #302, opened 2018-05-06
Stop exiting needlessly on plain "C" locales

jpouellet:
Without setting a specific *.UTF-8 locale, things default to C,
which uses a US-ASCII character encoding.

Default installs of at least OpenBSD (and likely other systems)
fall into this category, and asciinema refuses to run at all.

This is **ascii**nema after all... it should support ascii!

### DIFF
--- a/asciinema/__main__.py
+++ b/asciinema/__main__.py
@@ -50,8 +50,8 @@ def maybe_str(v):
 
 
 def main():
-    if locale.nl_langinfo(locale.CODESET).upper() != 'UTF-8':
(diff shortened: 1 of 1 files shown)


sickill:
That's a good observation sir :)

A little story about the UTF-8 requirement: it saved on a lot of complexity in asciinema-player to only support Unicode when interpreting input. Also, JSON files are supposed to be UTF-8 encoded, so adding this check makes sure we don't need to do any encoding conversion before saving to file, neither when loading it in the player.

Given any US-ASCII text is also a valid UTF-8 text I don't see why we should not support good old ASCII!

Thanks!

PR #324, opened 2018-10-01
Add configuration for which keys pause and step playback

lucywyman:
**What this changes** This adds configuration options for which characters will cause playback to pause and step through playback
**Why** I'm demoing [Bolt](https://github.com/puppetlabs/bolt) at Puppetize Live next week, and wanted a tool that would allow me to look like I'm typing while having a predictable demo. There's a similar tool that does this, 'Demo easel', but it only works on OSX and Windows. Since I didn't want to mess with a VM during my demos, hacking this together seemed like the easiest way to get a close approximation.

### DIFF
--- a/asciinema/__main__.py
+++ b/asciinema/__main__.py
@@ -27,7 +27,10 @@ def rec_command(args, config):
 
 
 def play_command(args, config):
-    return PlayCommand(args.filename, args.idle_time_limit, args.speed)
+    chars = {}
+    chars['pause'] = config.play_pause_chars
+    chars['step'] = config.play_step_chars
+    return PlayCommand(args.filename, args.idle_time_limit, args.speed, chars)
 
 
 def cat_command(args, config):
--- a/asciinema/commands/play.py
+++ b/asciinema/commands/play.py
@@ -5,17 +5,18 @@ import asciinema.asciicast as asciicast
 
 class PlayCommand(Command):
 
-    def __init__(self, filename, idle_time_limit, speed, player=None):
+    def __init__(self, filename, idle_time_limit, speed, chars, player=None):
         Command.__init__(self)
         self.filename = filename
         self.idle_time_limit = idle_time_limit
         self.speed = speed
+        self.chars = chars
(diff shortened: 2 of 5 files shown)


lucywyman:
I think a couple of apologies are in order for this PR:
1. I'm sorry for not discussing this before putting it up. It's something I need for a demo next week, so was going to do regardless and figured discussion on the PR was just as good as in IRC.
2. I'm sorry for my terrible tests! I couldn't find a good place for integration tests for configuration, but it's been a long time since I read python I may have missed something. Let me know if there are more / better tests I can write.
3. Lastly apologies for my terrible python. It's been a while :)

As previously stated I'm making these changes locally for a demo next week, and thought others might be able to use the work. That said I completely understand if this isn't in line with the direction of the project, or if it seems too rude to have just put this up. Happy to discuss further, and thanks for taking the time!

lucywyman:
Also worth noting: I've neglected to update docs. I don't plan to unless this is accepted into the project, so...will do later if applicable!

sickill:
Hey @lucywyman 👋 

I hope your demo went well :)
No need to be sorry for "terrible tests" or "terrible python" :) This is a very clean PR in my opinion.

Apologies for not addressing this sooner. Indeed one of the reason I hesitated, was that we didn't have proper discussion on how to approach this.

The use case you had (looking like typing the commands) could be implemented as "demo mode" feature. We could have additional `asciinema play --demo file.cast` option that would turn any keypress into a step (what's triggered today only by `.` key). I think this would be better than emulating demo demo via lots (all letters?) `step_chars`. Also, if you'd like to use arrow key which generates a sequence of characters (`\e[C`) instead of single char, the approach you implemented won't work.

I'm also working on #340 and #341 which both would benefit from customizable key bindings as well. These are for `rec` subcommand, and there they will need either a "prefix" key defined (like tmux or GNU screen) or accept shortcuts with control key, like `<C-p>`. 

So I'd like to cover customization of key bindings for all commands (`play` and `rec` as of now) in a more complete, general way. Given that I think I'm going to close this one.

PR #326, opened 2018-10-29
Add long_description, data_files and test_requires to setup.py

Crestwave:


sickill:
Thank you @Crestwave!

Issue #328, opened 2018-11-04
don't print "asciinema: press <ctrl-d> or type "exit" when you're done" when recording some command instead of shell


vanyakosmos:

sickill:
That's good idea 👍

PR #330, opened 2018-12-06
Change "how to stop recording" info message

vanyakosmos:
- prints "exit opened program when you're done" if command was specified, or  "press <ctrl-d> or type "exit" when you're done" otherwise
- add better .gitignore file from [github/gitignore](https://github.com/github/gitignore/blob/master/Python.gitignore) (+ old one). Added it because I needed to hide my .venv :(
- slight refactoring (sort import, wrap really long line)

this pull request will close issue #328

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,120 @@
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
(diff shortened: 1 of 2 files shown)


sickill:
Thanks @vanyakosmos 👍
(I've fixed the E402 error in develop already)

PR #339, opened 2019-03-15
Edit recorded asciicasts, delete or squash events, concatenate recordings

### LINKED ISSUE #338: Edit recorded asciicasts (cut and concatenate parts)
This is a feature request I started to work on. The idea is, the user edits the recorded asciicasts and changes the event types. Take for example this recording with edited event types,
```
[1, "o", "Hel"]
[2, "s", "l"]
[3, "s", "o"]
[4, "d", " whirled"]
[5, "o", " world"]
```
The extended event types are
- "s": squeeze with the previous event
- "d": delete this event

The result of the above example is
```
[1, "o", "Hello"]
[5, "o", " world"]
```
Editing would be invoked `asciinema edit --from part1.cast --from part2.cast result.cast`. Each input file is processed the way detailed above, then concatenated into a single cast file with the right offset.

adamschmideg:
Fixes #338 

Take one or more asciicast files and edit them manually. Change the event
types as follows,

- "d": delete, this event will be deleted
- "s": squash, this event will be squashed onto the previous one. 
  More consecutive lines can be squashed.

After the individual files are processed, they are concatenated into a single
recording. The times are adjusted, so the last event in the first file is 
immediately followed by the first event in the second file, and so on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,6 +233,21 @@ output (including all escape sequences) to a terminal immediately.
 `asciinema cat existing.cast >output.txt` gives the same result as recording via
 `asciinema rec --raw output.txt`.
 
+### `edit -f <edited-file-1> [-f <edited-file-2> ...] <result-filename>`
+
+__Process manually edited asciicasts and join them to a single recording.__
+
+Take one or more asciicast files and edit them manually. Change the event
+types (it's the second item and usually an "o") as follows,
+
+- "d": delete, this event will be deleted
+- "s": squash, this event will be squashed onto the previous one. 
+  More consecutive lines can be squashed.
+
+After the individual files are processed, they are concatenated into a single
+recording. The times are adjusted, so the last event in the first file is 
+immediately followed by the first event in the second file, and so on.
+
 ### `upload <filename>`
 
 __Upload recorded asciicast to asciinema.org site.__
--- a/asciinema/__main__.py
+++ b/asciinema/__main__.py
@@ -119,7 +119,7 @@ For help on a specific command run:
     parser_cat.set_defaults(func=cat_command)
 
     # create the parser for the "edit" command
-    parser_edit = subparsers.add_parser('edit', help='Edit one or more recordings and save the result to a new recording')
+    parser_edit = subparsers.add_parser('edit', help='Edit recordings and concatenate them')
     parser_edit.add_argument('filename', help='filename/path to save the edited recording to')
     parser_edit.add_argument('-f', '--from-file', help='file(s) to edit', action='append')
     parser_edit.set_defaults(func=edit_command)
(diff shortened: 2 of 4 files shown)


sickill:
Hey @adamschmideg! Thanks for taking a stab at this. I was thinking myself about adding some kind of automated editing support, and was even playing with few ideas. I guess I should have put the ideas out there already so we could discuss.

Anyway, I have following thoughts on the proposed changes:

I don't quite see a value for marking an event for deletion... If you (or some tool) want to delete an event, deleting the line seems like the simplest way to do it, no? :) 

Then, using event's type field for putting the operation to perform has few problems.

[The set of event types](https://github.com/asciinema/asciinema/blob/develop/doc/asciicast-v2.md#supported-event-types) is open for future extension (for example the work in #341 adds new `b` event), so we're risking conflicts here.

Also, the problem with "squash" action is that events of different types (`o`, `i`, `b`) can't be squashed together. For example:

```json
...
[0.900000, "i", "a"]
[1.000000, "o", "some text"]
[1.200000, "i", "b"]
[2.000000, "o", "more text"]
[2.500000, "b", ""]
[3.750000, "o", "more text"]
[4.500000, "b", ""]
...
```

You can't squash the 4th event `o` into the 3rd (`i` one), since they have different semantics.
To make this work, by squashing 4th event we would need to squash it into the last preceeding event of the same type (`o` here), which is 2nd event. But, that would only work for `o` events. With `i`, which represents single keystrokes, you can't merge 2 keys into 1 without losing information :) With `b`, which represent breakpoints, there's also no good squashing result.

Maybe we should rethink this? What are your thoughts?

adamschmideg:
Regarding deletion, :man_facepalming:. It's obvious you can simply delete the unwanted lines. I don't know how I overlooked it :) So I'd simply remove "d".

I understand squashing would work only for consecutive "o" events, events with an "i" or "b" in between wouldn't be squashed. The event type could be "so" to denote it's only for "o" events. Stdout events are the most frequent ones, so I still think it could be a useful editing tool.

I think there are not so many event types now, so we can avoid conflicts on a case by case basis.

To give some background to this PR, I'm working on a tool that [automatically generates a recording from a shell script](https://gitlab.com/offcode/asciinema-turk). It's like `asciinema rec -c` on steroids, well, that's the plan. My editing needs may be ad-hoc.

To recap, I'm happy to remove the "d" case and make an "so" event type work only for consecutive "o" events if you think it gives value.

sickill:
I believe attacking the general problem of "editing", on a broader scope, would be better. There's some discussion about this happening here: https://github.com/asciinema/asciinema/issues/385

Given this and my other concerns about the solution proposed in this PR that I shared earlier I'm going to close the PR. Thanks anyway!

PR #340, opened 2019-03-17
Pause/resume recording

sickill:
This one adds recording pausing/resuming functionality, that can be used to temporarily turn off capturing of stdout/stdin by pressing a hotkey. This is useful when you want to execute commands during the recording session that should not be captured (e.g. pasting secrets).

Similarly to #341, the hotkey for toggling pause is hard-coded at the moment to `ctrl+p` ("p" for "pause capture"), but it is very likely to conflict with many interactive applications (e.g. vim), so we may want to find a better one, or go with a prefix + key (like `ctrl+a <key>` in `screen`, `ctrl+b  <key>` in `tmux`).

It may be a good idea to show a notification that pause/resume actually happened after pressing a hotkey. Printing something to the terminal is not a good idea here, since it would potentially overwrite whatever was printed before (in fullscreen apps like vim), so I believe an external tool is the best choice. We could have a setting in a config file `notification_command` which would be invoked to notify about pause/resume (also about setting breakpoint in #341). You could then set it to show the notification in tmux's status bar (via `tmux display-message`) or show a desktop notification (`notify-send`, Growl, etc) or play a sound.

- [x] toggle pause during recording session via hotkey
- [x] stop time when paused
- [x] documentation
- [ ] notifications
- [ ] find good default hotkey, and/or hotkey prefix (like GNU screen, tmux)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,5 +13,5 @@ before_install:
   - pip install pycodestyle
 
 script:
-  - find . -name \*.py -exec pycodestyle --ignore=E501,E402 {} +
+  - find . -name \*.py -exec pycodestyle --ignore=E501,E402,E722 {} +
   - make test
--- a/README.md
+++ b/README.md
@@ -137,9 +137,9 @@ By running `asciinema rec [filename]` you start a new recording session. The
 command (process) that is recorded can be specified with `-c` option (see
 below), and defaults to `$SHELL` which is what you want in most cases.
 
-You can temporarily "mute" recording by pressing <kbd>Ctrl+P</kbd>. This is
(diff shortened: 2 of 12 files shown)


sickill:
Maybe it makes more sense to call this feature "recording session pausing", since usually when you mute, the time continues to run, but here the goal is to stop the time too, otherwise we would leave idle gaps.

sickill:
Let's tackle notifications and keyboard shortcuts in separate PRs, since these two are going to be used by other commands too.

sickill:
Default key combo for pause is now `C+\`

PR #341, opened 2019-03-17
Breakpoints

sickill:
This one adds breakpoints - locations in the recording which allow the player to auto-pause when one is encountered.

A breakpoint is new `"b"` event type in asciicast stream, with optional annotation (comment) as event data.

To add a breakpoint during recording session press a hotkey (specified in config file as `add_breakpoint_key`).

To add breakpoints to existing recording you add `"b"` events like this:

```
...
[1.000000, "o", "some text"]
[2.500000, "b", ""]
[3.750000, "o", "more text"]
[4.500000, "b", "chapter 5"]
...
```

- [x] support for breakpoints in asciicast v2 event stream
- [x] setting breakpoint during recording session via hotkey
- [x] documentation
- [x] ~~find good default hotkey, and/or hotkey prefix (like GNU screen, tmux)~~ no default for now
- [x] implement pause-on-breakpoint mode for `asciinema play` command (opt-in with `-b/--breakpoints` switch)
- [x] jump to next breakpoint during `asciinema play` with hotkey (`]`)
- [x] add test

### DIFF
--- a/README.md
+++ b/README.md
@@ -469,7 +469,7 @@ pause_key = p
 ; default: . (dot)
 step_key = s
 
-; Define hotkey for jumping to next breakpoint,
+; Define hotkey for jumping to the next breakpoint,
 ; default: ]
 next_breakpoint_key = b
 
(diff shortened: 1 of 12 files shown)


abernier:
@sickill what about giving an optional name breakpoints : a such breakpoint can then be considered like a "chapter" (like in youtube) ?

sickill:
Bringing this PR from the dead :rocket:

PR #350, opened 2019-04-09
fix typo

davidbradway:


sickill:
Thanks!

davidbradway:
Very nice tool. Thanks! Got it working on alpine Linux via Docker and pip

sickill:
Awesome! I'm not english native speaker so I'm always happy when someones finds these 👍 

Btw, there seems to be native Alpine package.

PR #357, opened 2019-06-02
edit upload .cast file url

v1xingyue:
Some http servers such as nginx, will return a 301 redirect , when you post to a url not ended with / . 

When Post To  /asciinema/api/asciicasts , 
It will return  : 
301 /asciinema/api/asciicasts/ ,

### DIFF
--- a/asciinema/api.py
+++ b/asciinema/api.py
@@ -27,7 +27,7 @@ class Api:
         return "{}/connect/{}".format(self.url, self.install_id)
 
     def upload_url(self):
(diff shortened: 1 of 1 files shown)


sickill:
Thanks, however, for the ones which don't redirect and don't handle path with and without trailing slash the same it could also break upload. asciinema-server handles the one without slash just fine. What kind of setup you have where this doesn't work today? I don't think Nginx will do the redirect when using `upstream` directive unless it's explicitly told so.

PR #368, opened 2019-09-04
Check playback size can fit in terminal

geisler:
Before playing a recorded file, check that the terminal that is running
the playback is at least as large as the recorded content.  This will
allow all the escape characters that change the location on the screen
to go to a legal spot and, therefore, everything should display
properly.

I've chosen to warn for a smaller size instead of causing an error in
case the playback doesn't use escape codes that are too large, but one
could argue that an error is legitimate here, too.

### DIFF
--- a/asciinema/asciicast/v1.py
+++ b/asciinema/asciicast/v1.py
@@ -34,6 +34,9 @@ class Asciicast:
     def events(self):
         return self.stdout_events()
 
+    def get_size(self):
+        return (int(self.v2_header['width']), int(self.v2_header['height']))
+
     def stdout_events(self):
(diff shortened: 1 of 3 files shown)


ku1ik:
Thanks. This seems useful but not always necessary (e.g. if there's no shell right prompt). Let's revisit this in 3.0 or 3.1. I'm currently rewriting asciinema in Rust (https://github.com/asciinema/asciinema/milestone/8), and this will allow use of [avt](https://github.com/asciinema/avt) to relatively reliably detect if there would be any explicit cursor jump (other than regular print) to illegal spot before starting the playback. This way we could warn only if playback would actually result in the illegal thing.

Issue #364, opened 2019-07-24
ValueError: Invalid header value b'Basic <ConsumerKey>', error when uploading the local cast


mochadwi:
I'm trying to upload the local cast, but below error is thrown instead.

Here's the error.
```
Traceback (most recent call last):
  File "/usr/local/Cellar/asciinema/2.0.2/libexec/bin/asciinema", line 11, in <module>
    load_entry_point('asciinema==2.0.2', 'console_scripts', 'asciinema')()
  File "/usr/local/Cellar/asciinema/2.0.2/libexec/lib/python3.7/site-packages/asciinema/__main__.py", line 131, in main
    code = command.execute()
  File "/usr/local/Cellar/asciinema/2.0.2/libexec/lib/python3.7/site-packages/asciinema/commands/upload.py", line 14, in execute
    result, warn = self.api.upload_asciicast(self.filename)
  File "/usr/local/Cellar/asciinema/2.0.2/libexec/lib/python3.7/site-packages/asciinema/api.py", line 40, in upload_asciicast
    password=self.install_id
  File "/usr/local/Cellar/asciinema/2.0.2/libexec/lib/python3.7/site-packages/asciinema/urllib_http_adapter.py", line 76, in post
    response = urlopen(request)
  File "/usr/local/Cellar/python/3.7.4/Frameworks/Python.framework/Versions/3.7/lib/python3.7/urllib/request.py", line 222, in urlopen
    return opener.open(url, data, timeout)
  File "/usr/local/Cellar/python/3.7.4/Frameworks/Python.framework/Versions/3.7/lib/python3.7/urllib/request.py", line 525, in open
    response = self._open(req, data)
  File "/usr/local/Cellar/python/3.7.4/Frameworks/Python.framework/Versions/3.7/lib/python3.7/urllib/request.py", line 543, in _open
    '_open', req)
  File "/usr/local/Cellar/python/3.7.4/Frameworks/Python.framework/Versions/3.7/lib/python3.7/urllib/request.py", line 503, in _call_chain
    result = func(*args)
  File "/usr/local/Cellar/python/3.7.4/Frameworks/Python.framework/Versions/3.7/lib/python3.7/urllib/request.py", line 1360, in https_open
    context=self._context, check_hostname=self._check_hostname)
  File "/usr/local/Cellar/python/3.7.4/Frameworks/Python.framework/Versions/3.7/lib/python3.7/urllib/request.py", line 1317, in do_open
[...]

PR #376, opened 2019-10-03
Make it possible to read the user var from the config

### LINKED ISSUE #362: [Feature Request] Make the USER env var configurable
Hi,
I use asciinema from different machines with different USER env vars.
I didn't want to permanently change the variable on my other machine and it would be great if this variable could come from the config file too.

This is the related line I guess: https://github.com/asciinema/asciinema/blob/develop/asciinema/commands/command.py#L10

Gerifield:
Fixes #362 

This small change make it possible to set the user part for the basic auth in the upload process without affecting the USER env var.

- On my laptop the USER env var contains a full email address which made it impossible to upload any records due to an exception in the basic auth. Similar to #364 
- I also wanted to upload some recordings to my main account which was set from a different laptop and I don't want to change/set the USER var for this.

### DIFF
--- a/asciinema/commands/command.py
+++ b/asciinema/commands/command.py
@@ -7,7 +7,7 @@ class Command:
 
     def __init__(self, args, config, env):
         self.quiet = False
-        self.api = Api(config.api_url, env.get("USER"), config.install_id)
+        self.api = Api(config.api_url, config.user, config.install_id)
 
     def print(self, text, file=sys.stdout, end="\n", force=False):
         if not self.quiet or force:
--- a/asciinema/config.py
+++ b/asciinema/config.py
@@ -85,6 +85,10 @@ class Config:
         )
 
     @property
+    def user(self):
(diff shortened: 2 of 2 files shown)


acolomaedu:
May be you doesn´t have enough privileges to read the file o var encironment, try to upload the file with sudo

Gerifield:
Are you sure it's the comment for me? My problem was a different issue not related to sudo.

sickill:
The first issue (invalid header value) has been already solved.

The `$USER` var is not used for identifying the asciinema account, the UUID (install ID) is instead. The user name is set in the basic auth header only for the purpose of saving it as a _temporary_ display name, for the asciicasts who are uploaded by users who don't authenticate on the website. In other words, it's temporary and doesn't really matter. If you had the same `$USER` value on multiple machines you can still use 2 separate asciinema.org accounts. And, if you had 2 different `$USER` values (on different machines) you can still use the same single asciinema.org account if you want.

So in the end your proposed doesn't really solve anything, therefore I'm going to close this.

PR #378, opened 2019-11-06
Add local file readers for compressed .gz and .bz2 files

tomdottom:
Adds the ability for `asciinema` to read from `.gz.` and `.bz2` compressed files.

```shell
asciinema play tests/demo.json.gz
asciinema play tests/demo.json.bz2
asciinema play tests/demo.cast.gz
asciinema play tests/demo.cast.bz2
```


asciinema/encodings.py
```diff
@@ -0,0 +1,52 @@
...
+)
+
+
+def search_function(encoding):
+    if encoding == "gzip":
+        return gzip_codec
+    return None
+
+
+codecs.register(search_function)
```
sickill, line 48: Looks this one is duplicate of the one below in `setup()`. I assume that's not intentional?

asciinema/encodings.py
```diff
@@ -0,0 +1,52 @@
...
+    decode=gzip_decode,
+    streamreader=GzipStreamReader,
+    streamwriter=GzipStreamWriter,
+    _is_text_encoding=False,
+)
+
+
+def search_function(encoding):
+    if encoding == "gzip":
+        return gzip_codec
```
sickill, line 44: Why do we need one for gzip but not for bz2? 🤔

asciinema/asciicast/__init__.py
```diff
@@ -77,6 +84,14 @@ def open_url(url):
 
         return utf8_reader(body, errors='strict')
 
+    _, encoding = mimetypes.guess_type(url)
+    if encoding == "gzip":
+        with codecs.open(url, mode="r", encoding="gzip") as fh:
+            return StringIO(fh.read().decode("utf-8"))
```
sickill, line 90: This requires loading whole recording into memory, which is incompatible with [streaming features introduced in 2.0](https://blog.asciinema.org/post/two-point-o/). I'd rather keep it consistent, and made gzip/bz2 support streaming.

Something like this (for gzip case) should work I think:

```python
# gzip
return gzip.open(url, mode='rt', encoding='utf-8')

# bz2
return bz2.open(url, mode='rt', encoding='utf-8')
```

https://docs.python.org/3/library/gzip.html
https://docs.python.org/3/library/bz2.html

sickill:
Closing as the code review concerns were not addressed. Feel free to re-open if you wish to pursue this further.

PR #398, opened 2020-03-21
Consume terminal answerbacks in asciinema cat

djpohly:
If the asciicast contains escape sequence queries like "CSI 6 n", the terminal will both echo and input its responses.  Use the raw() context manager that we already have to attempt to consume this "input."

There is, unfortunately, no way of finding out exactly when the terminal has finished its answerbacks.  This patch adds a 50ms wait, which should be overkill for a local terminal (in my tests, 3ms was usually enough).  For a remote terminal, this number becomes harder to estimate.

Technically this only needs to be done if whatever we're writing to isatty(), but keep it simple for now.

### DIFF
--- a/asciinema/commands/cat.py
+++ b/asciinema/commands/cat.py
@@ -1,6 +1,7 @@
 import sys
 
 from asciinema.commands.command import Command
+from asciinema.term import raw
 import asciinema.asciicast as asciicast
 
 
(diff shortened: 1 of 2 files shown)


sickill:
Good catch! This looks good 👌

sickill:
@djpohly To check how perceivable 50ms is, I changed it to 10ms and I can feel the difference actually. Not that it's a big deal, but I also think 50ms is an overkill, and I'd prefer to lower it. Given your tests showed ~3ms is enough, and 10ms feeling almost instant, do you agree that we can lower it to 10ms with reasonably high confidence it will be enough?

Remote terminal latency is a different story, and there we can't have any guarantees, so I'd rather not worry about it, at least not now (and from my knowledge most asciinema users record local, not remote sessions).

djpohly:
I wouldn't have a problem with it.  (I had the remote case for `asciinema cat` in mind since I was brainstorming options for remote teaching at that point.)  I can't think of a simple way to guarantee that the wait will be longer than one round-trip from the running program to the user's terminal and back without adding something clunky like an "End of stream, press enter" prompt.

sickill:
Thanks for the insight 👍

PR #405, opened 2020-04-10
Add "--show-timer" play option

hydrargyrum:
This option will show the current frame timer in a corner of the display.

This option can be useful when a user wants to edit a cast file:
- user plays the file with --show-timer and jots down interesting timestamps
- user knows where to insert/remove what they want in the cast file


ku1ik:
That's a clever use of `SC` (save cursor) / `RC` (restore cursor) sequences, but I'm afraid we can't have it. The problem with this approach is that it breaks assumptions of programs running in a terminal about the saved cursor position. For example, vim may use `SC` to save cursor position and want to do `RC` to jump back, but the timer stored different position in saved cursor "register", so vim won't be able to restore it correctly. We would need to restore saved cursor to its original value but we can't read that from terminal emulator.

I think a better way would be to implement this as a hook - you specify a hook command and it gets executed periodically with current stream time. This way you could for example use `tmux` status bar to display the time.

PR #422, opened 2020-11-27
Modernize asciinema packaging (use twine, github actions..)

gpotter2:
Asciinema is using a lot of deprecated stuff to package itself. This is why, for instance, https://pypi.org/project/asciinema/2.0.2/ isn't rendered and looks terrible. I also took the liberty to migrate from Travis CI to Github CI since it has a much better integration in Github, and allowed to split the jobs to separate pycodestyle. (also because Travis CI is currently doing all it can to kick OpenSource out..)

This PR:
- migrates from Travis CI to Github CI. Create a standalone job: "pycodestyle". The badge will need to be updated once this is merged.
- update the `Makefile` to use `twine`, as [you should](https://packaging.python.org/guides/distributing-packages-using-setuptools/).
- adds a `twine check` test. good practice
- use `license_file` in `setup.cfg`
- drop `tests_require`: it's not used since no test entry is specified in `setup.py` and is deprecated.

Job demo: https://github.com/gpotter2/asciinema/actions/runs/387286730. Please make sure that Github Actions are not disabled in the repo settings (they shouldn't by default). I'm pretty sure this needs to be merged once for the actions to start on future PRs.


sickill:
Awesome! Thanks @gpotter2 👍 Let's give it a go.

PR #434, opened 2021-07-15
[WIP] Windows Support

zurgeg:
A Windows support PR now that Windows has PTY.


votagus:
Why was this closed?

zurgeg:
> Why was this closed?

I think it ended up not working, I wouldn't know since it's been awhile. FYI though, you can use WSL then run `cmd` (or powershell or whatever) through WSL and it'll work fine

rollingmoai:
@votagus You can try this port for Windows: [Watfaq/PowerSession-rs](https://github.com/Watfaq/PowerSession-rs)

PR #444, opened 2021-11-04
Replace mentions of freenode with libera.chat

handlerug:
Partially solves https://github.com/asciinema/asciinema-server/issues/385.


sickill:
Thanks!

Btw, what's that "gamja" thing in the Libera.Chat web client URL?

handlerug:
It's the new webchat, which is called gamja. IMO it's better than KiwiIRC, but obviously it's your choice whether to use it or not :)

sickill:
Sounds good. Thanks for clarification!

PR #454, opened 2021-12-07
Feature: implement --begin-at flag for play command (fix #447)

### LINKED ISSUE #447: [Request] Implement startAt for asciinema cli
As noted in https://github.com/asciinema/asciinema-player/issues/156 it would be great to have a startAt flag that I could give the fractional timestamp from the log file in order to implement a search functionality.

I guess it'll be enough when the played skips the first n-lines from a local .cast file up to the occurence of the timestamp.

tomowang:
Add one new cli flag `--begin-at` for play command. This option accepts
one fractional as relative time and will skip till this time when playing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,7 @@ Available options:
 
 * `-i, --idle-time-limit=<sec>` - Limit replayed terminal inactivity to max `<sec>` seconds
 * `-s, --speed=<factor>` - Playback speed (can be fractional)
+* `-b, --begin-at=<ts>` - Begin at relative time when playing (can be fractional)
 
 > For the best playback experience it is recommended to run `asciinema play` in
 > a terminal of dimensions not smaller than the one used for recording, as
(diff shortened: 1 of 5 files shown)


fahuihu:
这是来自QQ邮箱的自动回复邮件。您好，我已收到您的邮件，谢谢对外经济贸易大学，信息学院，朱文强

sickill (changes requested):
Current implementation won't work reliably for all recordings because it ignores all print events that happened before given point in time. Thing is, terminal emulators are state machines and they build up their state (internal as well as visible one - screen lines) incrementally. For example, if you have a recording of a vim session and you would try playing it from the middle you need to first print everything that came before, otherwise the playback will produce garbage. So we need to first print everything up to given time in one shot ("catch up"), then continue with regular timed playback.

Also, we need to make it play nicely with `--out-fmt` and `--stream` option of `play` command. It would make sense to catch up for "o" (stdout) stream but not for "i" (stdin).

But I'm not sure if "catching up" is in line with your desired use case.

tomowang:
I see. As vim case, I really noticed that the output was broken when remove any leading lines.

PR #455, opened 2021-12-08
Replace nose with unittest

### LINKED ISSUE #388: The Nose unittest library has reached end-of-life
From https://nose.readthedocs.io/en/latest/:

> Nose has been in maintenance mode for the past several years and will likely cease without a new person/team to take over maintainership. New projects should consider using Nose2, py.test, or just plain unittest/unittest2.

Unless there is a strong requirement for Nose features, may I suggest converting the tests to [the standard Python unittest](https://docs.python.org/3/library/unittest.html) library?

stratakis:
This PR is quite extensive as it re-organizes the tests directory structure in order to make it compatible with unittest, however I hope the conversion is pretty straightforward. This way the tests can be run via the command line `python3 -m unittest discover` or also by invoking `pytest` directly should anyone desires. I've removed some unused code as well.

All the tests pass locally with this PR, I haven't checked the integration tests though.

Resolves #388


asciinema/tests/test_config.py
```diff
@@ -0,0 +1,212 @@
+import unittest
```
stratakis, line 1: This file was renamed and changed as well, so the diff is not representative unfortunately, it would be better to compare the current `tests/config_test.py` with it for a proper diff.

asciinema/tests/test_helper.py
```diff
@@ -0,0 +1,20 @@
+import sys
```
stratakis, line 1: This file was also moved and changed, so the diff is not representative, not sure why github is not showing in a similar manner as test_pty. It would be better to compare the current `tests/test_helper.py` with it for a proper diff.

sickill:
Thanks!

As you noticed @djds has been working on similar a thing in #445. There's been a lot of different changes there but hopefully we'll be able to look at them in separate smaller PRs (#456, #457, #458). I don't have a strong preference on whether to move the tests to a different dir or not, and I'm not sure what's the common practice in Python land these days.

Either way I'm happy to let you and @djds figure out which version of tests refactoring to use.

djds:
@stratakis `pytest` can be used as the runner for `unittest`, so I think it makes sense to go the `pytest` route which still allows for writing `unittest` tests. Working out how to get my original changes into `develop` so we don't duplicate effort too much, but it should be possible to merge these tests with some changes even if we use `pytest`. Been busy with `${dayjob}` unfortunately, but I should have some time over the holidays to get these things resolved.

sickill:
I agree with @djds, let's go with `pytest` 👍

djds:
@stratakis #445 has been merged, so now it should be easier to refactor some tests.

stratakis:
Simply running pytest should be sufficient. I could refactor the PR if anyone requires the tests to be also invoked through unittest, however I believe depending on pytest should be enough. Closing this and thanks @sickill and @djds for resolving the issue with #445 :)

Issue #116, opened 2015-06-30
Packaging of asciinema 2.1.0


sickill:
Opening this issue to discuss the best way for package maintainers to be notified about new asciinema releases.

We could create a mailing list for this, or use https://github.com/asciinema/asciinema/releases.atom like @jakubjedelsky suggested (https://github.com/asciinema/asciinema/issues/43#issuecomment-117053123). Alternatively I could be reopening this issue and pinging you here in the comments whenever new release is out. Any other ideas and/or preferences?

/cc @jakubjedelsky @zanchey @Dr-Terrible @larryv

UPDATE: So I just keep this issue open and comment when there's new release. Package maintainers can subscribe to this issue to get notified.

zanchey:
The atom feed is enough for me. SourceForge used to be the place for mailing lists, but I think they've gone a bit evil now. We have a mailman server if you want to use that.

Dr-Terrible:
The atom feed is enough for me too.
If by any chance I miss a release and users are complaining, feel free to poke me via GitHub's issues through my Gentoo overlay.

swills:
Personally, I'd prefer an email of some kind, either mailing list or github issue email, either is fine. Thanks!

larryv:
I'd also prefer email notifications (to minimize update lag), but I'd be fine with the atom feed.

sickill:
I have just released 1.2.0 /cc @larryv @swills @Dr-Terrible @zanchey @jakubjedelsky

Dr-Terrible:
@sickill Done! Thank you for the reminder.

v1.2.0 tested with both Go1.4.x and Go1.5.x (Gentoo/Linux x86-64), and everything works as expected, with all the unit tests passing. Ebuild just pushed in my overlay: Dr-Terrible/go-overlay@485968240b2f646c675b0630f452d28e99ca41cc.

sickill:
@Dr-Terrible wow, that was quick! Thanks!

sickill:
@youtux thanks for updating Homebrew package!

youtux:
@sickill you are welcome. Also, I use [sibbell.com](sibbel.com) to get updates via email every time a starred github project gets a new tag.

zanchey:
I've updated the Ubuntu packages; sorry for the delay.

sickill:
@zanchey thanks!

larryv:
[updated in MacPorts](https://trac.macports.org/changeset/148664/), only 3 months late 😓

[...]

PR #459, opened 2021-12-11
Recommend pipx installation, remove sudo

harkabeeparolus:
I'm suggesting some improvements to the installation instructions, as discussed in https://github.com/asciinema/asciinema/issues/116#issuecomment-983691890 — see this comment for full rationale.

Basically we should never tell users to pip install with `sudo`, and also it would be helpful to recommend using `pipx` for a more reliable installation if the user has it.

Please note: I think the web site should also be updated accordingly:
https://asciinema.org/docs/installation


sickill:
Awesome, thank you.

PR #476, opened 2022-02-21
[typing] Annotate tests

djds:


sickill (approved):
Nice, thanks!

That Alpine build failure, I assume the 403 that we see there is a random/temporary problem, right?

djds:
> Nice, thanks!
> 
> That Alpine build failure, I assume the 403 that we see there is a random/temporary problem, right?

Yes. It's a permissions issue with how secrets are shared (they're not) with forks. Builds pass in my own namespace: https://github.com/djds/asciinema/actions/runs/1878917732

PR #477, opened 2022-02-22
djds/pylint

djds:
- [recorder] Specify UTF-8 encoding
- [pty_] Avoid shadowning class name with copy() method argument
- [style] Add pylint exceptions

### DIFF
--- a/asciinema/asciicast/raw.py
+++ b/asciinema/asciicast/raw.py
@@ -6,7 +6,7 @@ from ..file_writer import file_writer
 
 
 class writer(file_writer):
-    def __init__(
+    def __init__(  # pylint: disable=too-many-arguments
         self,
         path_: str,
         metadata: Any = None,
@@ -32,5 +32,6 @@ class writer(file_writer):
     def write_stdin(self, ts: float, data: Any) -> None:
         pass
 
+    # pylint: disable=consider-using-with
     def _open_file(self) -> None:
         self.file = open(self.path, mode=self.mode, buffering=self.buffering)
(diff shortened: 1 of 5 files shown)


asciinema/asciicast/raw.py
```diff
@@ -6,7 +6,7 @@
 
 
 class writer(file_writer):
-    def __init__(
+    def __init__(  # pylint: disable=too-many-arguments
```
sickill, line 9: Btw, how are you finding these? Running `pylint` manually, or you have pylint integration in your editor?

sickill (approved):

asciinema/pty_.py
```diff
@@ -150,7 +150,7 @@ def copy(signal_fd: int) -> None:  # pylint: disable=too-many-branches
     os.waitpid(pid, 0)
 
 
-class signal_fd:
+class SignalFD:
```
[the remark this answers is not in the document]
djds: I would usually stick to `PascalCase` class names unless there's an external reason for doing differently. Either way the issue here wasn't the name itself, but rather that `positional argument to `copy` was shadowing the `signal_fd` class:
```python
     def copy(signal_fd: int) -> None:  # pylint: disable=too-many-branches
         fds = [pty_fd, tty_stdin_fd, signal_fd]
         stdin_fd = pty.STDIN_FILENO
```

asciinema/asciicast/raw.py
djds: Both, but https://github.com/dense-analysis/ale is always running in the background.

sickill: Got it, thanks. `ale` looks nice. Yeah, I love shellcheck, taught me better scripting practices.

asciinema/pty_.py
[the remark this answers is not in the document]
sickill: Right, I saw the name clash. Was just curious about your general suggestion about the context-managers. Thanks!

Issue #438, opened 2021-10-03
[bug] Module 'locale' has no attribute 'nl_langinfo'


hazu-89:
## System Information
Asciinema version: 2.1.0
Python version: 3.9.7
OS: Android 10 (on Termux 0.117)

## Bug Description
Command `asciinema` exits prematurely with the following traceback:

```
Traceback (most recent call last):
  File "/data/data/com.termux/files/usr/bin/asciinema", line 8, in <module>
    sys.exit(main())
  File "/data/data/com.termux/files/usr/lib/python3.9/site-packages/asciinema/__main__.py", line 29, in main
    if locale.nl_langinfo(locale.CODESET).upper() not in ['US-ASCII', 'UTF-8', 'UTF8']:
AttributeError: module 'locale' has no attribute 'nl_langinfo'
```

Relevant documentation:
> locale.nl_langinfo(option)
> 
>    Return some locale-specific information as a string. This function is not available on all systems, and the set of possible options might also vary across platforms. 
[Python 3.8 documentation](https://docs.python.org/3.8/library/locale.html#locale.nl_langinfo)

Alternative: Use [locale.getlocale()](https://docs.python.org/3.8/library/locale.html#locale.getlocale).

sickill:
I guess we could catch the above exception and fallback to `locale.getlocale`.

JordonRR:
sherlock-program${--sites SITE_name}

kolayne:
This seems fixed by #503, isn't it?

PR #503, opened 2022-06-02
[bug #438] Fallback to locale.getlocale()

djds:


sickill (approved):
Great, thanks.

Btw, what OS is this needed for?

djds:
> Great, thanks.
> 
> Btw, what OS is this needed for?

Based on the original bug report #438, it looks like Android, but it's simple enough to implement using the standard library, so I figured "why not?".

sickill:
Sure. I had that original bug report in mind too, just thought maybe you stumbled upon this somewhere else.

PR #523, opened 2022-09-17
Add shell completions by shtab

Freed-Wu:
```sh
asciinema --print-completion bash | sudo tee /usr/share/bash-completion/completions/asciinema
asciinema --print-completion zsh | sudo tee /usr/share/zsh/site-functions/_asciinema
asciinema --print-completion tcsh | sudo tee /etc/profile.d/asciinema.completion.csh
```

```sh
❯ asciinema <TAB>
asciinema commands
auth
cat
play
rec
upload
❯ asciinema play --<TAB>
option
--help              show this help message and exit
--idle-time-limit   limit idle time during playback to given number of seconds
--loop              loop loop loop loop
--out-fmt           select output format
--speed             set playback speed (can be fractional)
--stream            select stream to play
❯ asciinema play --out-fmt <TAB>
out_fmt
asciicast  raw
```

It should can replace
<https://github.com/zsh-users/zsh/blob/master/Completion/Unix/Command/_asciinema>.
Because this PR is generated from argparse automatically. But the latter is 
written manually.

I use some `type:ignore`. It is not elegant. Does any better method exist?

### DIFF
--- a/asciinema/__main__.py
+++ b/asciinema/__main__.py
@@ -2,6 +2,7 @@ import argparse
 import locale
 import os
 import sys
+from copy import copy
 from typing import Any, Optional
 
 from . import __version__, config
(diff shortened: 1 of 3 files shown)


fahuihu:
这是来自QQ邮箱的自动回复邮件。您好，我已收到您的邮件，谢谢对外经济贸易大学，信息学院，朱文强

Freed-Wu:
@sickill Any problem?

ku1ik:
The code looks good, but due to other priorities I haven't had a chance to deal with this PR earlier, sorry.

Currently I'm rewriting asciinema in Rust (https://github.com/asciinema/asciinema/milestone/8) and there we'll most likely use https://crates.io/crates/clap_complete

Thanks!

PR #547, opened 2023-04-27
Add `--suppress-output` flag to `record` command

DavidVentura:
Hi

This PR implements a new flag for the `record` command: `--suppress-output`.
This allows certain use-cases, like running recorded integration tests in CI without the commands' output polluting logs.

I've only added unit-tests, as I could not get the expected integration test to work in `asciinema` before my changes:

```
$ asciinema rec --overwrite -q  -c "echo 'this is a very nice message'" out.rec | wc -c
this is a very nice message
0
```
I would've expected `wc` to return `28`.

Which I'd have compared with
```
$ asciinema rec --overwrite -q --suppress-output  -c "echo 'this is a very nice message'" out.rec  | wc -c
0
$
```

### DIFF
--- a/asciinema/__init__.py
+++ b/asciinema/__init__.py
@@ -21,6 +21,7 @@ def record_asciicast(  # pylint: disable=too-many-arguments
     title: Optional[str] = None,
     command_env: Any = None,
     capture_env: Any = None,
+    suppress_output: Optional[bool] = False,
 ) -> None:
     record(
         path_,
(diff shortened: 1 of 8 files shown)


ku1ik:
Thanks David.

What you want to achieve can be done by running `asciinema rec` in a new terminal "session" with [setsid](https://linux.die.net/man/2/setsid) command:

```bash
setsid asciinema rec -q -c 'ping -c 5 1.1.1.1' demo.cast
```

[asciicast](https://asciinema.org/a/615308)

`setsid` is available on every Linux system so there should be nothing to install in CI environments for this to work.

Given this produces desired outcome I believe there's no need to modify the code.

Issue #549, opened 2023-05-14
Using kitty terminal the `kitty +kitten ssh machine@ip` causes the shell to freeze during an `asciinema rec` session when asking for password input


AkechiShiro:
**Describe the bug**
Using https://github.com/kovidgoyal/kitty as a terminal, running `asciinema rec test.cast` and then `kitty +kitten ssh`, if a password prompt (see the custom password prompt) : 
![image](https://github.com/asciinema/asciinema/assets/14914796/c77d79e5-e308-4e03-8469-fe9597e91823)
 is asked by the remote ssh server, the shell freezes up.

To be clear, what we see during an asciinema record session is that : 
- Running `kitty +kitten ssh root@server.ssh`, the shell freezes up.
- Only running `kill -9` on the asciinema process allows getting back a working shell, CTRL+C does not work, nor any other signal for that matter.

**To Reproduce**
1. Launch `kitty` terminal
2. `asciinema rec test.cast`
3. `kitty +kitten ssh root@test.machine`
4. See a/b cases
  4a. If a password input is asked, the shell will freeze, this happens probably because kitty uses a custom way to handle password input for ssh (read additional context).
  4b. If no password prompt is asked, the ssh access works all fine, and there is no freeze, the recording is all good as well.

**Expected behavior**
In case a password is asked, the shell should not freeze, but the password input should be asked by kitty and then ssh login should be made, so the user can record their session.
**Versions:**
 - OS: ArchLinux Linux 6.2.10-arch1-1 
 - asciinema cli: 2.2.0

**Additional context**
This issue seems very specific to kitty terminal emulator, and more specifically to the use of https://sw.kovidgoyal.net/kitty/kittens/ssh/
I suppose the issue is due to a conflict between how the kitty native ssh password input, askpass handles the terminal.

[According to this documentation part about askpass](https://sw.kovidgoyal.net/kitty/kittens/ssh/#opt-kitten-ssh.askpass)
[...]

PR #556, opened 2023-06-20
Fix **write(2)** to pty master fd synchronously causing deadlock

Low-power:
**asciinema(1)** currently **write(2)**s to pty master fd synchronously, which could be blocking if the internal pty input buffer became full; while the programs running under the pty may also be blocking on writing to pty slave fd, because the pty output buffer can also become full at same time. Once this happens, the entire terminal session hangs; the only way to get rid of it is to kill the asciinema processes via `SIGKILL`.

The following steps can be used to reproduce the bug:

* Start a terminal recording
* Run `cat` in the pseudo terminal that's being recorded
* Prepare large amount of text and put it in X11 primary selection (or clipboard if that's preferred); how about 2000 lines of `w` repeating 80 times for each line?
```sh
i=0; while [ $i -lt 2000 ]; do echo wwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwww; i=$((i+1)); done | xclip -verbose -selection primary -in -target text/plain
```
* Paste the text into terminal that you previously runs **cat(1)** in
* **asciinema(1)** will likely be hanging at this time

The fix is to use **select(2)** in function `copy` to check whether the pty master fd is immediately writable, before trying to **write(2)** anything to it.

### DIFF
--- a/asciinema/pty_.py
+++ b/asciinema/pty_.py
@@ -37,6 +37,7 @@ def record(
     prefix_key = key_bindings.get("prefix")
     pause_key = key_bindings.get("pause")
     add_marker_key = key_bindings.get("add_marker")
+    input_data = bytes()
 
     def set_pty_size() -> None:
         cols, rows = get_tty_size()
@@ -54,6 +55,7 @@ def record(
             writer.write_stdout(time.perf_counter() - start_time, data)
 
     def handle_stdin_read(data: Any) -> None:
+        nonlocal input_data
         nonlocal pause_time
         nonlocal start_time
         nonlocal prefix_mode
@@ -84,10 +86,7 @@ def record(
 
             return
 
-        remaining_data = memoryview(data)
-        while remaining_data:
-            n = os.write(pty_fd, remaining_data)
-            remaining_data = remaining_data[n:]
+        input_data += memoryview(data)
 
         # save stdin unless paused or data is OSC response (e.g. \x1b]11;?\x07)
         if not pause_time and not (
@@ -100,14 +99,21 @@ def record(
             writer.write_stdin(time.perf_counter() - start_time, data)
 
     def copy(signal_fd: int) -> None:  # pylint: disable=too-many-branches
-        fds = [pty_fd, tty_stdin_fd, signal_fd]
+        nonlocal input_data
+
+        crfds = [pty_fd, tty_stdin_fd, signal_fd]
 
         while True:
+            if len(input_data) > 0:
+                cwfds = [pty_fd]
+            else:
+                cwfds = []
+
             try:
-                rfds, _, _ = select.select(fds, [], [])
+                rfds, wfds, _ = select.select(crfds, cwfds, [])
             except KeyboardInterrupt:
-                if tty_stdin_fd in fds:
-                    fds.remove(tty_stdin_fd)
+                if tty_stdin_fd in crfds:
+                    crfds.remove(tty_stdin_fd)
 
                 break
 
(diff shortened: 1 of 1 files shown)


Low-power:
## Some **truss(1)** logs showing the deadlock

**truss(1)** output for **asciinema(1)** process that's handing pseudo terminal I/O:
```
30904: write(4,"wwwwwwwwwwwwwwwwwwwwwwwwwwwwwwww"...,1024) = 1024 (0x400)
30904: _umtx_op(0x806e14014,UMTX_OP_SEM2_WAKE,0x0,0x0,0x0) = 0 (0x0)
30904: _umtx_op(0x806e14014,UMTX_OP_SEM2_WAIT,0x0,0x0,0x0) = 0 (0x0)
30904: _umtx_op(0x801e162b8,UMTX_OP_NWAKE_PRIVATE,0x1,0x0,0x0) = 0 (0x0)
30904: _umtx_op(0x8007f5008,UMTX_OP_WAIT_UINT_PRIVATE,0x0,0x18,0x7fffdfffd0e8) = 0 (0x0)
30904: write(9,"\0\0\^D \M^@\^C]q\0(G@\\\M-@\M-'"...,1060) = 1060 (0x424)
30904: read(0,"wwwwwwwwwwwwwwwwwwwwwwwwwwwwwwww"...,1024) = 1024 (0x400)
30904: write(11,"wwwwwwwwwwwwwwwwwwwwwwwwwwwwwwww"...,1024) = 768 (0x300)
30904: SIGNAL 9 (SIGKILL)
```

**truss(1)** output for **cat(1)** process:
```
31743: read(0,"wwwwwwwwwwwwwwwwwwwwwwwwwwwwwwww"...,4096) = 81 (0x51)
31743: write(1,"wwwwwwwwwwwwwwwwwwwwwwwwwwwwwwww"...,81) = 81 (0x51)
31743: read(0,"wwwwwwwwwwwwwwwwwwwwwwwwwwwwwwww"...,4096) = 81 (0x51)
31743: write(1,"wwwwwwwwwwwwwwwwwwwwwwwwwwwwwwww"...,81) = 81 (0x51)
31743: read(0,"wwwwwwwwwwwwwwwwwwwwwwwwwwwwwwww"...,4096) = 81 (0x51)
31743: write(1,"wwwwwwwwwwwwwwwwwwwwwwwwwwwwwwww"...,81) = 81 (0x51)
31743: read(0,"wwwwwwwwwwwwwwwwwwwwwwwwwwwwwwww"...,4096) = 81 (0x51)
31743: write(1,"wwwwwwwwwwwwwwwwwwwwwwwwwwwwwwww"...,81) = 1 (0x1)
31743: SIGNAL 9 (SIGKILL)
31743: process killed, signal = 9
```

As from these outputs, the last **write(2)** call hangs in both processes, until they received `SIGKILL`.

sickill:
Great, this problem is actually something we debugged recently in https://github.com/kovidgoyal/kitty/issues/6272 and we have open issue for https://github.com/asciinema/asciinema/issues/549

I'll take a look at your patch shortly. Thanks!

AkechiShiro:
@sickill When you get time to look at the patch in depth, could you let me know if this patch could fix the bug report I created ?

sickill:
Made a small fix to this https://github.com/asciinema/asciinema/commit/1514591628f6d121fb886ce2f5073336f8c466e8

sickill:
@AkechiShiro I tested on your bug repro instructions and it seems the patch fixes it indeed :tada:

sickill:
I'll release new version shortly.

AkechiShiro:
Thanks a lot !! @sickill @Low-power

AkechiShiro:
@sickill all the others terminal recorders must be affected by this race condition as well, I don't know if this bug should be reported to them as well, I'll let you handle that, as my issue is finally fixed.
Will give it a try when a new release is made of asciinema

sickill:
@AkechiShiro just released 2.3.0: https://pypi.org/project/asciinema/2.3.0/

PR #558, opened 2023-06-27
Make `cat` subcommand be capable to actually concatenate multiple recordings

Low-power:
Despite its name, the `cat` subcommand accepts only 1 input file, and can't really be used to concatenate multiple files.

Making it a true `cat` would be useful at least to myself.

### DIFF
--- a/asciinema/__main__.py
+++ b/asciinema/__main__.py
@@ -222,10 +222,10 @@ For help on a specific command run:
 
     # create the parser for the `cat` command
     parser_cat = subparsers.add_parser(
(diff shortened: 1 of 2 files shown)


sickill:
Thanks. This is nice improvement.

Btw, I now think that when I implement `cat` subcommand I meant it more as "cat which prints the contents of a file", rather than "cat which concatenates". So in fact more apt name would have been something like `asciinema dump <filename>` or `asciinema stdout <filename>`. That's probably why your expectation wasn't met :)

PR #562, opened 2023-07-21
fast forward by N frames (default 10)

maho:
- quite ugly way to fast-forward. Definitively could be improved, but is better than nothing.


ku1ik:
The code looks good, not ugly at all IMO :)

However, I'm not sure if usability of this would be high - 10 frames (or any N higher than 1) can fast-forward you into an unpredictable place in the future in recording's timeline, and there's no way go back (rewind) then. So if you overshoot you need stop and start from the beginning. Feels like frustrating experience to me. Implementing rewind by N frames (or rewind in general), which I believe would make fast-forward viable, is rather non-trivial task and that's not planned at the moment.

Also, I'm working on Rust rewrite (https://github.com/asciinema/asciinema/milestone/8), and not planning any new features in 2.x line.

Thanks!

Issue #369, opened 2019-09-04
“Broken pipe” errors when piping inside asciinema


klmr:
I’ve noticed the following behaviour in asciinema 2.0.2:

```
$ for i in {1..30000}; do echo $i; done >large-file
$ asciinema rec
asciinema: recording asciicast to /var/folders/dr/rp46tjqs5qzfj_ghjpxqrttr0000gn/T/tmprapla7oa-ascii.cast
asciinema: press <ctrl-d> or type "exit" when you're done
$ cat large-file | head
1
2
3
4
5
6
7
8
9
10
cat: stdout: Broken pipe
```

This happens whenever the first program in a pipeline receives a SIGPIPE from the second command, and the data to be piped exceeds a certain length (in my terminal, in the above, using `{1..20000}` is not reproducibly enough to trigger the behaviour, but it *sometimes* is). The behaviour is the same for a wide range of programs, not just `cat`.

(macOS 10.14.6, bash 4.4.19 with iTerm 3.3.2 as well as Terminal.app)

karlkfi:
I see this pretty repeatably when reading from urandom:

```
SUFFIX=$(< /dev/urandom tr -dc '[:digit:]' | head -c4)
tr: write error: Broken pipe
tr: write error
```

sickill:
I cannot reproduce the problem on either Linux or macOS anymore. I believe this got fixed in either 2.1.0 or 2.2.0. Feel free to re-open if it still doesn't work for you. Thx!

klmr:
Unfortunately I am still able to reproduce this issue with asciinema 2.2.0 — on both macOS 10.14.6 as well as Ubuntu 18.04.

Just to check whether something in my shell configuration might be interfering, I even launched it with a clean environment, but the issue persists:

```sh
env -i LANG=C.UTF-8 HOME=$HOME $(which asciinema) rec
```

Low-power:
I can reproduce this issue on FreeBSD with latest commit (e4fb2d6)

Low-power:
The simplest test case I came up is:

```
$ yes | true
yes: stdout: Broken pipe
```

This command line would normally produce no output.

### Some **truss(1)** logs

Outside asciinema:
```
$ truss -f yes | true
...
83196: write(1,"y\ny\ny\ny\ny\ny\ny\ny\ny\ny\ny"...,4096) ERR#32 'Broken pipe'
83196: SIGNAL 13 (SIGPIPE)
83196: process killed, signal = 13
```

[...]

PR #578, opened 2023-10-22
Re-enable default signal handler for SIGPIPE in child process

ku1ik:
Fix for #369.

asciinema parent process doesn't set handler for SIGPIPE, which by default sets SIG_IGN for this signal in a child process. This explicitly sets the handler in the child process to SIG_DFL.

This is something script and mosh do right before exec-ing the command in a child process:

script - https://github.com/util-linux/util-linux/blob/f3fd64111c28de1ecd47f7cd519120ffa531800d/term-utils/script.c#L987
mosh - https://github.com/mobile-shell/mosh/blob/1105d481bb9143dad43adf768f58da7b029fd39c/src/frontend/mosh-server.cc#L540-L546

Many thanks to @Low-power for figuring this out :clap:

### DIFF
--- a/asciinema/pty_.py
+++ b/asciinema/pty_.py
@@ -167,6 +167,7 @@ def record(
     pid, pty_fd = pty.fork()
 
     if pid == pty.CHILD:
+        signal.signal(signal.SIGPIPE, signal.SIG_DFL)
         os.execvpe(command[0], command, env)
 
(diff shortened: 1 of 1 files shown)


ku1ik:
More background on the core issue: https://github.com/python/cpython/issues/45993 - that one is about subprocess/exec but the same rules apply for pty fork we use.

ku1ik:
Released v2.4.0 with the fix: https://github.com/asciinema/asciinema/releases/tag/v2.4.0

PR #580, opened 2023-11-02
New impl of add-marker for `rec` command

ku1ik:


cmprmsd:
Markers are great! I just can't add labeled markers from command line programmatically. That would be cool for automation. :)

PR #589, opened 2023-11-04
Accept ISO-8859-1 as a locale

h3xx:
I was unable to get locale.nl_langinfo to print 'US-ASCII', however, in my testing on Linux, most combinations of LC_{CTYPE,*}=en_US returned 'ISO-8859-1'.

**Note about the supposed Rust rewrite**:  Even though the rewrite in Rust has yet to manifest its source code publicly, much less officially release, in keeping with spirit of the GPL, I still wish to share my changes back upstream.

### DIFF
--- a/asciinema/__main__.py
+++ b/asciinema/__main__.py
@@ -24,7 +24,7 @@ def valid_encoding() -> bool:
     if loc is None:
         return False
     else:
-        return loc.upper() in ("US-ASCII", "UTF-8", "UTF8")
+        return loc.upper() in ("ISO-8859-1", "US-ASCII", "UTF-8", "UTF8")
 
 
 def positive_int(value: str) -> int:


ku1ik:
Is ISO-8859-1 a subset of UTF-8?

h3xx:
> Is ISO-8859-1 a subset of UTF-8?

Short answer: No.

Longer answer: [ISO-8859-1](https://en.wikipedia.org/wiki/ISO/IEC_8859-1) aka latin-1 was the default terminal encoding for a long time. It is a superset of US-ASCII. Both latin-1 and US-ASCII are 8-bit encodings.

There do exist characters in latin-1 that don't directly map to UTF-8 codepoints, however, there are libraries -- including Python3's string library -- that will translate between them.

```python
utf8_string = latin1_string.decode('latin1').encode('utf8', 'ignore')
```

Programs and terminals that are in UTF-8 mode require handling ASCII control characters; "\e" (character `1b`) IS part of ASCII but ISN'T part of UTF-8.

h3xx:
I was unable to get `locale.nl_langinfo` to give back the string `'US-ASCII'` -- what locale settings would this give `'US-ASCII'` as the result?

PR #601, opened 2024-01-06
Replace mio with select in the recorder, add signal handling

### LINKED ISSUE #598: Recording broken on macOS
```sh
$ target/release/asciinema rec demo.cast
Error: Invalid argument (os error 22)
```

Seems to be related to https://github.com/tokio-rs/mio/issues/1377

ku1ik:
This fixes recording on macOS, where mio can't poll /dev/tty via kqueue.

Fixes #598

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -89,14 +89,12 @@ dependencies = [
  "anyhow",
  "clap",
  "config",
- "mio",
  "nix",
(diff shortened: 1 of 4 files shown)


ku1ik:
Signal handling need to be ported from the previous (mio) version.

PR #611, opened 2024-02-07
`stream` command

ku1ik:
This adds `asciinema stream` command, which allows viewing a live stream of a terminal session in a browser.

Start the stream with:

```
$ asciinema stream
::: Streaming session started, web server listening on http://127.0.0.1:8080
::: Press <ctrl+d> or type 'exit' to end
$
exit
::: Streaming session ended

```

View the stream at http://127.0.0.1:8080

---

This a an improved version of the live terminal streaming initially implemented in https://github.com/asciinema/tst (which was a prototype of this functionality). This PR includes the "local server" mode. The "forward to remote asciinema server" mode (available in tst) will be implemented in a future PR.


phiresky:
this is great! would it be possible to add an option to write/stream to a normal jsonl file simultaneously so you still have the recording?

PR #616, opened 2024-02-12
Bugfix: handle SIGALRM and SIGCHLD signals. Fixes #613

### LINKED ISSUE #613: Recording wont exit when using P10k's instant prompt
**Describe the bug**
asciinema rec wont terminate when exiting zsh shell if using p10k-instant-prompt.

**To Reproduce**
1. ascinema rec -c "zsh"
2. ...
3. exit
4. asciinema keeps recording

**Expected behavior**
Stop recording

**Versions:**
 - OS: Ubuntu 22.04
 - asciinema cli: asciinema 3.0.0-beta.2

Pytness:
Fixes https://github.com/asciinema/asciinema/issues/613.


src/pty.rs
```diff
@@ -2,6 +2,7 @@ use crate::io::set_non_blocking;
 use crate::tty::{Tty, TtySize};
 use anyhow::{bail, Result};
 use nix::errno::Errno;
+use nix::libc::{SIGALRM, SIGCHLD};
```
ku1ik, line 5: We already import other signals from `signal_hook::consts` (below), so let's add those there, ok?

ku1ik (changes requested):
Thanks for looking into this. It looks good. Let's just consolidate the constant imports.

ku1ik:
Thanks!

PR #634, opened 2024-05-23
feat: Make asciinema work with shebangs

iulian-birlica:
I would like to explore the idea of having asciinema working with shebangs, thus having asciicasts work by uttering:
```bash
./file.cast
```

I wanted to have some example code, in order to start the discussion.

The file would have to be made executable after it is created. I think this should be an argument to `asciinema rec`, something like `-e -- make executable cast`.

I think supporting arguments like `-s speed` and the rest is technically doable, but may require some command-line magic and maybe some changes to `asciinema play`, or even an entire new `play` mode.

### DIFF
--- a/src/asciicast.rs
+++ b/src/asciicast.rs
@@ -49,7 +49,11 @@ pub fn open_from_path<S: AsRef<Path>>(path: S) -> Result<Asciicast<'static>> {
 
 pub fn open<'a, R: BufRead + 'a>(reader: R) -> Result<Asciicast<'a>> {
     let mut lines = reader.lines();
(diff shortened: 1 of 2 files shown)


ku1ik:
Unfortunately we can't just add a shebang line to the cast file as it would be a breaking change. The format, https://docs.asciinema.org/manual/asciicast/v2/, doesn't support anything like this, and there's a whole ecosystem, which would break if the CLI started producing cast files with a new line like the one you propose. Small subset of things that use cast files is here: https://docs.asciinema.org/integrations/

iulian-birlica:
There isn't any v3 on the roadmap as well, right?

ku1ik:
Not at the moment.

PR #637, opened 2024-06-09
Ref/flake

Dich0tomy:
This PR refactors the recently added flake:

1. Added shell.nix and default.nix for users that don't use flakes and use Nix2
2. Removed flake-utils and used flake-parts instead, it's way less volatile and has a couple of nice features that could come in handy later (e.g. debugging, really nice modularization)
3. Changed up a few things to comply with best practices and conventions - like attribute names (`nativeBuildInputs` vs `packages` for `mkShell`), name conventions (`pname`, `name`, `version` for mkDerivation), etc.
4. Made the source of the derivation reproducible when built locally (`src = ./.` relies on the current folder name, which if changed can trigger cache misses)
5. The `default.nix` and `shell.nix` are `callPackage`d, which automatically fills their arguments with `nixpkgs` packages and allows for easy overriding
6. Split the package's dependencies into correct categories, in this case only 3 matter:
	- `nativeBuildInputs` - dependencies which are only needed in the build process and nowhere else (like build tools).
	- `buildInputs` - dependencies which will be needed by runtime later - static libraries linked into the library, dynamic libraries and such.
	- `nativeCheckInputs` - native dependencies for tests, this one has a separate category because one may wish to disable checks for the package which would also make the derivation not pull in the unnecessary dependencies.

	For more info on that matter see - https://nixos.org/manual/nixpkgs/stable/#ssec-stdenv-dependencies

As for the modularization - if you'd like the flake could also be split into actual modules, which would make the flake [look something like this](https://github.com/Dich0tomy/Dire/blob/trunk/flake.nix) and then [subsequent modules something like this](https://github.com/Dich0tomy/Dire/tree/trunk/nix).

As for maintenance of the package - changes are required if you add Rust libraries that depend on system installed libraries as well or add any kind of resources that would need to be installed into the system later (like man docs, shell completions, icons) - I'm happy to assist with that.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -9,20 +9,6 @@
   darwin,
   python3,
 }: let
-  testDeps = [
-    python3
-  ];
-
-  buildDeps = rust:
-    [
-      rust
-    ]
-    ++ (lib.optionals stdenv.isDarwin [
-      libiconv
-      darwin.apple_sdk.frameworks.Foundation
-    ])
-    ++ testDeps;
-
   mkPackage = rust:
     (makeRustPlatform {
       cargo = rust;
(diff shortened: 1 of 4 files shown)


ku1ik (changes requested):
Thanks. Quite a lot of changes, half of them I don't yet fully understand :sweat_smile: 

One thing I already noticed is that `nix develop .#msrv` doesn't use MSRV from Cargo.toml anymore:

```
marcin@puter ~/p/a/asciinema (Dich0tomy-ref/flake)> nix develop .#msrv
[marcin@puter asciinema]$ rustc --version
rustc 1.77.2 (25ef9e3d8 2024-04-09)
```

That should be 1.70.

Dich0tomy:
My bad, fixed!

> Quite a lot of changes, half of them I don't yet fully understand 😅

I will be more than happy to explain the parts you are unsure about!

ku1ik:
Lovely, thanks. I think I'm kinda getting most of it now :sweat_smile:

PR #639, opened 2024-07-01
Use POST when obtaining new stream metadata, use GET when specific stream requested

ku1ik:


ku1ik:
Server support added in https://github.com/asciinema/asciinema-server/commit/fbbb2ce35272d516ce2a6debfbc04c9a7c60a149

PR #642, opened 2024-07-18
feat: read REQUESTS_CA_BUNDLE env to maybe override root CA

fenollp:


fenollp:
https://github.com/asciinema/asciinema/pull/643 is probably a better option.

PR #643, opened 2024-07-19
fix(reqwest): use rustls-tls-native-roots to profite OS level custom CA certs

hydezhao:
Add add `rustls-tls-native-roots` feature to reqwest so that the binary can use CA certificates from OS's keystore.

It solves the problem that we cannot upload cast to a self-hosted server with a private CA certificate.


hydezhao:
I updated the PR to use `rustls-tls-native-roots` following discussions on https://github.com/asciinema/agg/pull/75

ku1ik:
Nice, thanks!

PR #650, opened 2024-09-19
Add support for play -r.

gnachman:
The new `-r` or `--resize` option causes asciinema to send `CSI 8 ; <width> ; <height> t` for `r` commands in the Asciicast. This causes the terminal window to resize as it did during the recording. This commit also sets the initial size. Doing so causes recordings of interactive programs to render properly.

Resizing is off by default for backward compatibility.

I see this [was discussed](https://github.com/asciinema/asciinema/issues/198) previously but as far as I can tell it was closed without being implemented.


ku1ik:
This is nice addition, thanks!

PR #652, opened 2024-10-05
fix: avoid closing pty master fd twice

kxxt:
The pty master fd is closed twice, once when dropping OwnedFd in `fn exec` and once when dropping File constructed with unsafe code in `fn copy`.

This causes IO Safety violation[^1] and should be avoided. It becomes a crash in debug build[^2].

[^1]: https://github.com/rust-lang/rust/commit/1ba00d9cb2fcfef464b6a188fa3a7543c66eecaa
[^2]: https://github.com/zed-industries/zed/issues/12114


ku1ik:
Good catch, thanks!

PR #653, opened 2024-10-05
Use safe kill from nix crate

kxxt:
Previously the code uses unsafe kill from libc and on the first look it appears that it forgets to check the result.

I think you mean to ignore the errors that might occur when killing the child, right? This commit makes the intent more clear with less unsafe code.


ku1ik:
Yes, exactly, I meant this piece to ignore the result, as there really wouldn't be much we could do here to handle the error case anyway. Thanks for the improvement!

PR #654, opened 2024-10-05
Avoid hardcoding 5 for errno EIO

kxxt:
We could use libc::EIO to improve readability of the code.


ku1ik:
Good improvement, thanks!

PR #676, opened 2025-04-10
Add gruvbox dark theme for player

hupfdule:


ku1ik:
The built-in themes are not really used when streaming in local mode (this is the only case where the player is used here), since we use terminal's original color theme. And this would be overwritten when we upgrade the bundled player in the future.

PR #677, opened 2025-04-28
Support unix sockets output

andrewschreiber:
Changes are optimized for current use case

May not want to add a terminal PID to the .cast event format, but it's key for me ingesting multiple terminals on a single socket.

### DIFF
--- a/src/asciicast.rs
+++ b/src/asciicast.rs
@@ -30,6 +30,7 @@ pub struct Header {
     pub command: Option<String>,
     pub title: Option<String>,
     pub env: Option<HashMap<String, String>>,
+    #[allow(dead_code)]
     pub child_pid: Option<u32>,
 }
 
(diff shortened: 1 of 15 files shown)


ku1ik:
Hey,

This seems like very specific to your own use case. Frankly none of it is something I'm interested in getting in, and let me explain why:

- adding additional fields to asciicast header - the file format is described ([here](https://docs.asciinema.org/manual/asciicast/v2/)) and pretty much fixed,
- adding additional, 4th element to event array - similarly, this is incompatible with the asciicast format and would break a lot of tooling (including asciinema server and player), also not sure why you need it on every event given that you added that to the header, and each connection handler (I assume you built) should have this info in its context,
- the unix socket output - interesting, but I don't think that's worth maintaining given a niche use case

You can use `socat` to connect asciinema's output to a unix socket, for example like this:

```
mkfifo /tmp/myfifo
asciinema rec /tmp/myfifo
```

In a separate terminal forward from fifo to the socket:

```
socat -u OPEN:/tmp/myfifo,rdonly UNIX-CONNECT:/tmp/mysocket
```

Good luck with whatever you're building!

PR #682, opened 2025-05-30
fix: typo in 'asciinema' in the authentication message

mustafamohsen:
The unauthenticated streaming attempt prints out an erroneous message:

> "this CLI hasn't been authenticated with {server_hostname} - run • `ascinema auth` first"


ku1ik:
Good catch, thanks!

Issue #680, opened 2025-05-27
[Feature Request] add cast name to final message and JSON header


tacman:
By the time I've finished recording, I sometimes forget the cast name.  Can you add it to the final message?

Before:
```
exit
::: asciinema session ended
```

After:
```
exit
::: asciinema session ended, casts/do-something-cool.cast saved with 45 lines
```

Cyb3rC3lt:
I'd like to second this idea please. For someone who pauses their VM and may have days to weeks of console lines scrolling back to see what cast I am currently recording to would be great. I personally don't need the number of lines but the cast location would be great.

moritzdietz:
Maybe something along the lines of this?!

```
diff --git i/src/cmd/session.rs w/src/cmd/session.rs
index a47f8a3..2c12ab1 100644
--- i/src/cmd/session.rs
+++ w/src/cmd/session.rs
@@ -184,7 +184,12 @@ impl cli::Session {
             debug!("shutdown complete");
         });

-        status::info!("asciinema session ended");
+        if let Some(path) = self.output_file.as_ref() {
+            status::info!("Recorded to {}", path);
+            status::info!("asciinema session ended");
+        } else {
+            status::info!("asciinema session ended");
+        }

         if !self.return_ || exit_status == 0 {
             Ok(ExitCode::from(0))
```

This prints the same info like when you begin a recording.
```
❯ ./target/release/asciinema rec ../asdasdas --overwrite
::: asciinema session started
::: Recording to ../asdasdas
::: Press <ctrl+d> or type 'exit' to end
```
and then
```
::: Recorded to ../asdasdas
::: asciinema session ended
```

moritzdietz:
> I'd like the filename in the header record as well.

I would suggest to use an environment variable you can set already without the need to alter the code.

You can use the option `--rec-env` and pass your own environment variable with whatever you want.

```
--rec-env <VARS>           Comma-separated list of env vars to capture [default: SHELL]
```

```Bash
$ export FOO="bar"
$ asciinema rec --rec-env FOO demo.cast

# < do your thing >
```

Then you can see the var in the header:
```Bash
$ jq -s '.[0]' test.cast
{
  "version": 3,
  "term": {
    "cols": 195,
    "rows": 53,
    "type": "tmux-256color",
    "version": "tmux next-3.6"
  },
  "timestamp": 1749218069,
  "env": {
    "FOO": "bar"
  }
}
```

tacman:
[...]

PR #683, opened 2025-06-06
Add filename of cast file to session end output

moritzdietz:
As discussed in #680 this adds some additional information to the output generated after asciinema exits.

Feel free to reject this PR if this do not want to add or update this part of the code.
But if you like the general idea but this needs changes, let me know.

### DIFF
--- a/src/cmd/session.rs
+++ b/src/cmd/session.rs
@@ -152,7 +152,12 @@ impl cli::Session {
             .await?
         };
 
-        status::info!("asciinema session ended");
+        if let Some(path) = self.output_file.as_ref() {
+            status::info!("asciinema session ended");
+            status::info!("Recorded to {}", path);
+        } else {
+            status::info!("asciinema session ended");
+        }
         shutdown_token.cancel();
 
         if let Some(task) = server {


moritzdietz:
Ok, I fixed it and resolved the merge. Although I noticed that current develop builds successfully, yet there is an issue where if you start a recording the process hangs and you can't exit asciinema.

ku1ik:
> I noticed that current develop builds successfully, yet there is an issue where if you start a recording the process hangs and you can't exit asciinema.

Oh, interesting. I assume that's on macOS? Could be this one #598

moritzdietz:
Yes, it's on macOS! Seems likely, although it's not throwing an error. 

Here is a recording:

https://github.com/user-attachments/assets/9fffbd46-2556-4f21-a4c3-49c9eac6d061

ku1ik:
That's the same issue. It just doesn't show the error now, and instead switches to headless mode automatically. It's not ideal behavior I guess. Anyway, no need for a separate issue.

ku1ik:
@moritzdietz the macOS recording is fixed now.

PR #698, opened 2025-09-25
chore: update dependencies

reneleonhardt:
### Chores
- Update cargo dependencies, dockerfile, flake and github-actions
- Let dependabot update cargo and docker
- Dependencies require MSRV 1.82

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,6 +1,13 @@
----
 version: 2
 updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "docker"
+    directory: "/"
+    schedule:
+      interval: "weekly"
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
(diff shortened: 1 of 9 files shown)


ku1ik:
Thanks. But what's the point of all of this? What problem are you trying to solve?

ku1ik:
> Is dependabot working? nix-quick-install-action@v33 should've been updated already.

It opened this PR https://github.com/asciinema/asciinema/pull/700 2 days ago, so seems to be working.

> Interesting that x86_64-apple-darwin is working with use-cross: false... but maybe switching to a native runner macos-15-intel would be better?

Good observation. I haven't ever tried to run this macOS intel binary, but also haven't got feedback it doesn't work. So it either works or nobody tried yet.

> Are windows builds possible now with Rust?

Yes and no. Rust would allow it, yeah, but Windows doesn't have UNIX-style PTY, which we use on Linux and macOS. For Windows we'd need an alternative implementation of session capture based on ConPTY, but I'm not yet ready to open this can of worms...

> edition 2024 isn't working yet

I'll try to make it work. I planned to do it soon anyway.

PR #722, opened 2026-02-05
allow playing a cast from stdin

fopina:
Using same code as convert uses for `INPUT`, allow playing a cast from stdin.

Personally, quite useful for longer casts that I have gzipped:
`zcat my.cast.gz | asciinema play -`

### DIFF
--- a/src/cmd/play.rs
+++ b/src/cmd/play.rs
@@ -12,7 +12,11 @@ impl cli::Play {
         let config = Config::new(None)?;
         let speed = self.speed.or(config.playback.speed).unwrap_or(1.0);
         let idle_time_limit = self.idle_time_limit.or(config.playback.idle_time_limit);
-        let path = util::get_local_path(&self.file)?;
+        let path: Box<dyn AsRef<std::path::Path>> = if self.file == "-" {
+            Box::new(std::path::Path::new("/dev/stdin"))
+        } else {
+            util::get_local_path(&self.file)?
+        };
         let keys = get_key_bindings(&config.playback)?;
         let runtime = Runtime::new()?;
 
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -320,7 +320,7 @@ pub struct Record {
 
 #[derive(Debug, Args)]
 pub struct Play {
-    /// The path to an asciicast file or HTTP(S) URL to play back. HTTP(S) URLs allow playing recordings directly from the web without need for manual downloading. Supported formats include asciicast v1, v2, and v3.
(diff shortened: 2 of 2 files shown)


ku1ik:
Thanks. This is something I want to support.

The reason why it hasn't been implement in 3.x yet is here: https://github.com/asciinema/asciinema/blob/f93ee80bb903eaaf8dda5f6fcd7b27895d4b1177/src/player.rs#L165-L166

The thing is, it reads whole input into memory (due to some difficulties I had with using non-async Iterator in the async player code). Fundamentally there's no reason the whole file needs to be loaded into memory first, so I'd like to find a way to get it working.

Until the the above problem is solved, piping recordings to stdin will have 2 potential issues:

1. if what you're piping is "infinite" (e.g. piping from a named pipe to which `asciinema rec` writes in real-time) it will never start playing and just hang
2. if you have very long and big gzipped session it may cause OOM

I'll try to look into this problem shortly, keep your PR open, and merge it once we can safely ship this.

If you have ideas for solving the iterator vs async context problem I'm all ears.

fopina:
Thanks for the feedback! My rust knowledge is 0, I’m only to cargo build and copy paste. I knew convert read from stdin so I looked at it and seemed like a good clean carry over 😄 

Nevertheless I understand the issue you mentioned, I’ll see if I can map it to the actual code and look for solutions (as I guess that also applies to convert as it currently is)👍 

Quick question : the issue already exists given an actual big file in disk, correct? Not related to the particular way of internal handling /dev/stdin?

fopina:
@ku1ik just created #724 to suggest something Claude suggested after digging through git history to understand why you were using `.collect` 😄 
Let me know how that sounds - it does work at least!

src/cmd/play.rs
```diff
@@ -12,7 +12,11 @@ impl cli::Play {
         let config = Config::new(None)?;
         let speed = self.speed.or(config.playback.speed).unwrap_or(1.0);
         let idle_time_limit = self.idle_time_limit.or(config.playback.idle_time_limit);
-        let path = util::get_local_path(&self.file)?;
+        let path: Box<dyn AsRef<std::path::Path>> = if self.file == "-" {
```
ku1ik, line 15: Let's move the check for `"-"` into `util::get_local_path` itself, since it's the one mapping paths/URLs/anything into valid local paths.

ku1ik (changes requested):

ku1ik:
Actually this is fine already 👍

ku1ik:
Thanks!

fopina:
Sorry I meant to pick it up later tonight!

ku1ik:
No worries. In the end I was on the fence about this, and it works well so all good. Thx!

PR #723, opened 2026-02-07
Add support for attaching to existing shell sessions

usman-heracles:
## Summary

Adds the ability to record or stream an already-running shell session instead of always spawning a new one. This is implemented via two new CLI options:

- `--pty <PATH>` - Attach directly to a PTY device (e.g., `/dev/pts/5` on Linux or `/dev/ttys005` on macOS)
- `--pid <PID>` - Resolve a process ID to its controlling PTY and attach to it

## Use Cases

- Record a long-running shell session that was started before asciinema
- Capture terminal output from a process you're already interacting with
- Non-intrusively record/stream a session without changing the user's workflow

## Usage Examples

```bash
# Find your shell's PTY
tty
# /dev/ttys005

# In another terminal, attach and record
asciinema rec --pty /dev/ttys005 output.cast

# Or use the PID directly
asciinema rec --pid 12345 output.cast
```

## Implementation Details

- When attached, asciinema captures output only (read-only mode)
- The original shell continues running normally
- Press Ctrl+C to stop recording; the shell is not affected
- Works with `rec`, `stream`, and `session` commands

## Test plan

- [ ] Verify `--pty` option works with Linux PTY paths
- [ ] Verify `--pty` option works with macOS PTY paths  
- [ ] Verify `--pid` option resolves to correct PTY on Linux
- [ ] Verify `--pid` option resolves to correct PTY on macOS
- [ ] Verify recording captures output correctly
- [ ] Verify Ctrl+C stops recording without affecting the shell
- [ ] Verify conflicting options (--pty with --command) are rejected

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -308,6 +308,32 @@ pub struct Record {
     #[arg(long, value_name = "PATH", help = "Log file path", long_help)]
     pub log_file: Option<PathBuf>,
 
+    /// Attach to an existing PTY device instead of spawning a new shell. Specify the path to the
+    /// PTY device (e.g., /dev/pts/5 on Linux or /dev/ttys005 on macOS). Use the `tty` command in
+    /// the target shell to find its PTY path. This allows recording an already-running shell session.
+    /// Cannot be used together with --command or --pid.
+    #[arg(
+        long = "pty",
+        value_name = "PATH",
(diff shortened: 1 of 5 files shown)


usman-heracles:
@codex review

PR #724, opened 2026-02-09
allow `play` to stream events

fopina:
This change removes the memory bottleneck in the play command by streaming events from disk instead of loading them all into memory.

This is greatly assisted by Claude but after a long chat around the existing `.collect` and its motivation.

Changes:
- Replace tokio::spawn with tokio::task::spawn_blocking in emit_session_events()
- Remove .collect() that was loading all events into a Vec
- Use blocking_send() instead of async send() for the channel

Benefits:
- Memory usage reduced
- Instant playback start (no pre-loading delay)
- Support for arbitrarily large recordings
- Foundation for stdin streaming support (cat file.cast | asciinema play -)

Technical details:
The original code used .collect() because tokio::spawn requires 'static lifetime. By using spawn_blocking instead, we can stream the iterator directly since blocking tasks don't have the same lifetime constraints. File I/O is blocking anyway, so this is the correct approach.

Testing:
- Generated different large files and compared Max RSS between develop version and my branch
- ```./ gtime -v ./asciinema-original play --speed 10000 size_100.cast
        Maximum resident set size (kbytes): 221712
        ./ gtime -v ./asciinema-new play --speed 10000 size_100.cast
        Maximum resident set size (kbytes): 111120
        ./ gtime -v ./asciinema-original play --speed 10000 size_500.cast
        Maximum resident set size (kbytes): 924512
        ./ gtime -v ./asciinema-new play --speed 10000 size_500.cast
        Maximum resident set size (kbytes): 508768
        ./ gtime -v ./asciinema-original play --speed 10000 size_2000.cast
        Maximum resident set size (kbytes): 2003312
        ./ gtime -v ./asciinema-new play --speed 10000 size_2000.cast
        Maximum resident set size (kbytes): 701328
   ```
- As this is Max RSS it, the increased memory is likely from the OS and not the binary but it still shows a much friendlier scaling of memory usage for the input file :smile:

### DIFF
--- a/src/asciicast.rs
+++ b/src/asciicast.rs
@@ -19,7 +19,7 @@ pub use v3::V3Encoder;
 pub struct Asciicast<'a> {
     pub version: Version,
     pub header: Header,
-    pub events: Box<dyn Iterator<Item = Result<Event>> + 'a>,
+    pub events: Box<dyn Iterator<Item = Result<Event>> + Send + 'a>,
 }
 
 #[derive(Debug, Copy, Clone, PartialEq)]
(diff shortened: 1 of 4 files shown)


ku1ik:
This looks like a good solution. Thanks!

PR #744, opened 2026-05-16
Implement seeking and stepping backwards

cizra:
Patch by GPT-5.5


ku1ik:
Thanks, but this is not something I plan to maintain at this moment.

PR #752, opened 2026-09-05
Pin GitHub Actions to commit SHAs

lpmi-13:
## What and why

This commit uses commit SHA hashes for the action pins instead of mutable version tags. This eliminates an accidental overwrite of the upstream tag as an issue, and since both renovate and dependabot speak commit sha natively now, there's no reason not to do this.

## Prior discussion

N/A

## How it was tested

N/A, involves CI

## AI assistance

I asked copilot to do this work, and my comments on it are below in the diff.

## Checklist

- [x] I fully understand the changes I'm proposing and can answer questions about them
- [x] I wrote the description above in my own words
- [x] The test suite passes locally

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09 # v5.1.0
 
       - name: Create the release
         env:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -53,16 +53,16 @@ jobs:
       CARGO: cargo
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09 # v5.1.0
 
       - name: Install Rust toolchain
-        uses: dtolnay/rust-toolchain@stable
+        uses: dtolnay/rust-toolchain@6bed0761d98439e5a578e2877258200ad565ba87 # stable
```
lpmi-13, line 59: this is potentially the only change we don't want to do, if the preferred intent is to track `stable` automatically, though it's possible that dependabot could do this automatically as well. Happy to revert in any case if this change isn't desirable.

lpmi-13:
I would be very happy to add in a dependabot config to automatically bump these with a 2 week cooldown period (or whatever other period might be useful) either in this PR or a followon.

